PR #8, opened 2024-01-02
Improved Text Retrieval & Split-Screen Window

sudhanshu19102003:
## Changes Made
- Modified the window opening behavior to split-screen.
- Replaced nvim.current.eval("getreg('\"')") with nvim.current.buffer[:] for retrieving the code.
> [!TIP]
>nvim.current.eval("getreg('\"')") :
> It only contains the text from the most recent yank, delete, change, or put operation, if there were no changes it didn't give any output
**Before**
Screenshot 2024-01-02 151549
**After**
Screenshot 2024-01-02 151714

### DIFF
--- a/rplugin/python3/plugin.py
+++ b/rplugin/python3/plugin.py
@@ -45,7 +45,7 @@ class CopilotChatPlugin(object):
             prompt = prompts.EXPLAIN_SHORTCUT
 
         # Get code from the unnamed register
-        code = text = self.nvim.current.buffer[:]
+        code = self.nvim.current.buffer[:]
         file_type = self.nvim.eval("expand('%')").split(".")[-1]
         # Check if we're already in a chat buffer
         if self.nvim.eval("getbufvar(bufnr(), '&buftype')") != "nofile":
--- a/rplugin/python3/plugin.py
+++ b/rplugin/python3/plugin.py
@@ -45,7 +45,7 @@ class CopilotChatPlugin(object):
             prompt = prompts.EXPLAIN_SHORTCUT
 
         # Get code from the unnamed register
-        code = self.nvim.eval("getreg('\"')")
+        code = text = self.nvim.current.buffer[:]
         file_type = self.nvim.eval("expand('%')").split(".")[-1]
         # Check if we're already in a chat buffer
         if self.nvim.eval("getbufvar(bufnr(), '&buftype')") != "nofile":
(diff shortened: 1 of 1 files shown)


jellydn:
I've implemented a similar approach on my fork https://github.com/jellydn/CopilotChat.nvim/commit/97fcd701bd6ee38a997b25ca170270acf5198b3d 

I think the view mode/option should be configured.

```lua
return {
  -- Import the copilot plugin
  { import = "lazyvim.plugins.extras.coding.copilot" },
  {
    dir = "~/Projects/research/CopilotChat.nvim",
    branch = "canary",
    opts = {
       mode = "split"
    },
    build = function()
      vim.defer_fn(function()
        vim.cmd("UpdateRemotePlugins")
        vim.notify("CopilotChat - Updated remote plugins. Please restart Neovim.")
      end, 3000)
    end,
    event = "VeryLazy",
    keys = {
      { "<leader>cce", "<cmd>CopilotChatExplain<cr>", desc = "CopilotChat - Explain code" },
      { "<leader>cct", "<cmd>CopilotChatTests<cr>", desc = "CopilotChat - Generate tests" },
    },
  },
}
```

sudhanshu19102003:
Ya that's great but we still have to change the 
 `nvim.current.eval("getreg('"')") :` to `nvim.current.buffer[:] `

jellydn:
> Ya that's great but we still have to change the `nvim.current.eval("getreg('"')") :` to `nvim.current.buffer[:] `

I am not sure about it. Let's see @gptlang view on your PR.

gptlang:
I completely missed this PR

gptlang:
After looking at it a bit more, I agree with @jellydn's approach of being able to configure it (even though I do split pretty much 99% of the time)

rplugin/python3/plugin.py
```diff
@@ -45,12 +45,16 @@ def copilotChat(self, args: list[str]):
             prompt = prompts.EXPLAIN_SHORTCUT
 
         # Get code from the unnamed register
-        code = self.nvim.eval("getreg('\"')")
+        code = self.nvim.current.buffer[:]
```
gptlang, line 48: Would this pass the whole file in the buffer in or just the visible part?

sudhanshu19102003: The whole file which is opened currently

PR #9, opened 2024-01-03
Use `buf_set_lines` & `buf_set_text` methods instead of `nvim.current.line`

ziontee113:
This PR uses `buf_set_lines` & `buf_set_text` methods to append new tokens to the end of the chat buffer, no matter which window is active and no matter the user's current cursor position.

This makes moving the cursor / changing window mid generation possible without new tokens spewing out at the current cursor position.

### DIFF
--- a/rplugin/python3/plugin.py
+++ b/rplugin/python3/plugin.py
@@ -65,10 +65,7 @@ class CopilotChatPlugin(object):
 ### Copilot
 
 """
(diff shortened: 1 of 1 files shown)


jellydn:
Nice job 👏 
@gptlang I was about to report the issue then I saw this PR. This is working nicely. Could you take a look? Thanks.

gptlang (approved):
Thanks!

PR #60, opened 2024-02-20
fix: add check for temperature if empty string

banjocat:
This was causing an error on a fresh install.

100% understand if rather have a different fix!

Thanks for an awesome plugin.


jellydn (approved):
LGTM! Thanks @banjocat

PR #79, opened 2024-02-26
docs: input for items with args

nisalVD:
Hi, 
I don't know if this the correct way to use this, but when i tried it made more sense to have a space after these commands for input.

calling CopilotChatBuffer with no args throw

```
E471: Argument Required
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -53,7 +53,7 @@ return {
     end,
     event = "VeryLazy",
     keys = {
-      { "<leader>ccb", "<cmd>CopilotChatBuffer<cr>", desc = "CopilotChat - Chat with current buffer" },
+      { "<leader>ccb", "<cmd>CopilotChatBuffer ", desc = "CopilotChat - Chat with current buffer" },
       { "<leader>cce", "<cmd>CopilotChatExplain<cr>", desc = "CopilotChat - Explain code" },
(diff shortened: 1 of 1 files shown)


jellydn (approved):
Thanks @nisalVD LGTM!

jellydn:
@all-contributors add @nisalVD for doc.

README.md
```diff
@@ -53,7 +53,7 @@ return {
     end,
     event = "VeryLazy",
     keys = {
-      { "<leader>ccb", "<cmd>CopilotChatBuffer<cr>", desc = "CopilotChat - Chat with current buffer" },
```
MikaelElkiaer, line 56: It seems that `<cmd>` should be changed to `:` - at least I got an error in neovim version `NVIM v0.10.0-dev-2457+ga4b444252`.

nisalVD: yea you are right

nisalVD (commented):
it should be that

README.md
MikaelElkiaer: Sorry for just dropping a comment instead of a PR, hoping it was better to leave a comment than nothing. :)

nisalVD: Thanks @MikaelElkiaer made a new PR with that change https://github.com/CopilotC-Nvim/CopilotChat.nvim/pull/88

adrien-barret:
don't know it's juste a command problem. I have the same code error and I launched the command via `:` in nvim

PR #83, opened 2024-02-27
Refactor the plugin to be lua-based

deathbeam:
- Rewrite copilot class to lua
- Implement vsplit chat in lua
- Move prompts to lua
- Move commands to lua

## Changes
- Plugin now depends on https://github.com/nvim-lua/plenary.nvim (mainly for curl module)
- Removed vlog code replaced it with plenary.log as it was copied from there anyway
- Copilot jobs are now properly terminated on window close/window loss and new chat prompts
- .ask and .open can accept config as second param to override some parts of the config per command (for example `chat.ask('Hello', { window = { layout = 'float' } })`)
- - Prompts can now be defined as tables (name = prompt still works too) and when defined as table you can specify config just for that prompt (similar as chat.ask/open). mapping is also supported,that automatically creates keybiding for that prompt
- Merged some of InPlace chat features with VSplit chat features so now vsplit chat:
  - Supports different layouts (horizontal,vertical split and floating window)
  - Now you can type to the chat and submit a prompt
  - Now you can make code replacement from the chat
  - Now you can trigger completion for user and system prompts when pressig tab after /. You can also use / syntax to enter shortcuts to your prompts directly or to use different system prompt
  - All the keybindings on chat are now configurable
  - There is now help as virtual text in the split chat
  - Spinner is now as virtual text at end of written text by copilot
  - Separators are now different due to support for submitting prompts (its just ---, e.g markdown horizontal rule, configurable)
- Did not reimplemented InPlace chat in lua as vsplit chat with some of the in place interactive features is better design imo so did not wanted to spend energy on it
- Did not reimplemented requesting authentication in Lua, it only gets the token now. Imo its better to let plugins like copilot.vim developed by tpope and microsoft to handle the initial auth. But if this is not desired I can move the python part as well as its not hard
- Removed `CopilotChatVisual` and `CopilotChatBuffer`, this can be handled with selectors instead
- By default `CopilotChat` now behaves like CopilotChatVisual, with configurable selector for unnamed buffer. It looked like unnamed buffer was workaround with some selection issues anyway but with new selector system I can just pass whatever data I want to the commands

## TODO
- [ ] For proxy support, this is needed: https://github.com/nvim-lua/plenary.nvim/pull/559
- [ ] Delete rest of the python code? Or finish rewriting in place then delete
- [ ] Check for curl availability with health check
- [ ] Add folds logic from python, maybe? Not sure if this is even needed
- [ ] As said in changes part, finish rewriting the authentication request if needed
- [ ] Properly get token file path, atm it only supports Linux (easy fix)
- [ ] Update README and stuff
- [ ] Add token count support to extra_info, something like this called from lua:
```python
@pynvim.function("CopilotChatTiktokenLen", sync=true)
def CopilotChatTiktokenLen(self, args)
   import tiktoken
   input = args[0]
   model = args[1]
   return len(tiktoken.encoding_for_model(model).encode(input))
end
```

## New config example

```lua
local chat = require('CopilotChat')
local select = require('CopilotChat.select')
local prompts = require('CopilotChat.prompts')

-- This is just default config + some custom prompts showing mappings
chat.setup {
  system_prompt = prompts.COPILOT_INSTRUCTIONS,
  model = 'gpt-4',
  temperature = 0.1,
  debug = false,
  clear_chat_on_new_prompt = false,
  disable_extra_info = true,
  name = 'CopilotChat',
  separator = '---',
  prompts = {
    FixDiagnostic = {
        prompt = 'Please assist with the following diagnostic issue in file:',
        selection = select.diagnostics,
        mapping = '<leader>ar',
    },
    Explain = {
        prompt = prompts.USER_EXPLAIN, -- You can also just use prompt value straght from prompts module
        mapping = '<leader>ae',
    },
    Tests = {
        prompt = '/USER_TESTS', -- You can use references like this, works even outside of completion
        mapping = '<leader>at',
    },
    Documentation = {
        prompt = 'Add documentation comments to the selected code.',
        mapping = '<leader>ad',
    },
    Fix = {
        system_prompt = prompts.COPILOT_FIX, -- Or override system prompt like this
        prompt = 'Propose a fix for the problems in the selected code.',
        mapping = '<leader>af',
    },
    Optimize = {
        prompt = '/COPILOT_DEVELOPER Optimize the selected code to improve performance and readablilty.', -- Or use system prompt reference in prompt
        mapping = '<leader>ao',
    },
    Simplify = {
        prompt = 'Simplify the selected code and improve readablilty',
        mapping = '<leader>as',
    },

  },
  selection = function()
    return select.visual() or select.line()
  end,
  window = {
    layout = 'vertical',
    width = 0.8,
    height = 0.6,
    border = 'single',
    title = 'Copilot Chat',
  },
  mappings = {
    close = 'q',
    reset = '<C-l>',
    complete = '<Tab>',
    submit_prompt = '<CR>',
    submit_code = '<C-y>',
  },
}

vim.keymap.set({ 'n', 'v' }, '<leader>aa', chat.toggle, { desc = 'AI Toggle' })
vim.keymap.set({ 'n', 'v' }, '<leader>ax', chat.reset, { desc = 'AI Reset' })
```

### Floating window
image

### Completion
image


deathbeam:
Added some more examples to prompts with overriding and references

gptlang:
I'll do some experiments and see if it works

gptlang:
Notes:
- https://github.com/mlua-rs/mlua - Generate module bindings. Cannot bind structs.
- Use the tiktoken lib from https://github.com/dqbd/tiktoken/blob/main/core/src/lib.rs because it strips out the PyO3 bindings and is purely native (meaning that we only need to bind the functions we need and removes extra dependencies)

gptlang:
Looks like I need to refactor the library to use functional programming to pass around state. 

Reference: https://github.com/mlua-rs/mlua/issues/130

gptlang:
WIP here: https://github.com/gptlang/tiktoken.lua

jellydn (approved):
Great. Amazing work @deathbeam Let me merge to canary and test. Thanks.

PR #85, opened 2024-02-28
Add newline before CopilotChat: if first token is not alphanumeric

deathbeam:
This is needed so syntax highlighting for stuff like code blocks or lists do not breaks when copilot replies those right away


jellydn (approved):
LGTM! Thanks.

PR #87, opened 2024-02-28
Add migration guide to README.md

deathbeam:
Did not updated rest of the README yet but at least for migration this should be enough.

### DIFF
--- a/MIGRATION.md
+++ b/MIGRATION.md
@@ -1,0 +1,54 @@
+# Migration guide after Copilot Chat rewrite to Lua
+
+## Prerequisities
+
+Ensure you have the following plugins installed:
+
+- [plenary.nvim](https://github.com/nvim-lua/plenary.nvim)
+- [copilot.vim](https://github.com/github/copilot.vim) (recommended) or [copilot.lua](https://github.com/zbirenbaum/copilot.lua)
+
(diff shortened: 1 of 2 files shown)


jellydn (commented):
Thanks, @deathbeam Could you add a working example with the canary branch? Then it would be super useful if someone want to take atry. What do you think?

deathbeam:
> Thanks, @deathbeam Could you add a working example with the canary branch? Then it would be super useful if someone want to take atry. What do you think?

By working example you mean like full config with something like lazy.nvim? As I dont use plugin managers so im not confident in making the working example actually work :d

jellydn:
If so @deathbeam Could you share how to setup it on your end? I will translate to the plugin managers :)

deathbeam:
> If so @deathbeam Could you share how to setup it on your end? I will translate to the plugin managers :)

Sure, I have submodule for
- [copilot.vim](https://github.com/github/copilot.vim)
- [plenary.nvim](https://github.com/nvim-lua/plenary.nvim)
- [CopilotChat.nvim(https://github.com/CopilotC-Nvim/CopilotChat.nvim) on branch `canary`

And then this is how I configure it for myself atm (from linked line to end of file):
https://github.com/deathbeam/dotfiles/blob/master/nvim/.config/nvim/lua/config/copilot.lua#L14

PR #92, opened 2024-02-28
Fix issues with visual selection and folds, add set_debug

deathbeam:
- Visual selection now uses also '< and '> as fallback when in command line mode
- Remove vim.print that I forgot when setting folds
- Add set_debug function that updates M.config.debug and updates logger
- Properly update prompts in case reference in prompt references another reference


jellydn (approved):
LGTM!

PR #96, opened 2024-02-29
Make code_actions accept config as parameter

deathbeam:


jellydn (approved):
Great. Thanks @deathbeam LGTM!

PR #98, opened 2024-02-29
feat: add git diff selection and commit message generation

deathbeam:
This commit introduces the ability to select git diff content and generate commit messages following the commitizen convention. It adds two new selection methods, `Commit` and `CommitStaged`, to the `CopilotChat` module. These methods return the current git diff, with `CommitStaged` specifically returning staged changes. The `gitdiff` function in the `select` module is also updated to support this feature. This enhancement improves the utility of the chatbot in a version control context.

**This message was generated by this change :d**


jellydn (approved):
LGTM! Thanks @deathbeam

PR #99, opened 2024-02-29
feat(CopilotChat): Improve line separation and prompt update logic, add diff view

deathbeam:
Now submit_code and submit_prompt properly try to find first match from the end so you dont need to cursor over specific blocks. This is also prerequisite for showing current diff as float or in corner somewhere.

### DIFF
--- a/lua/CopilotChat/copilot.lua
+++ b/lua/CopilotChat/copilot.lua
@@ -70,7 +70,7 @@ local function generate_request(history, selection, filetype, system_prompt, mod
   if selection ~= '' then
     -- Insert the active selection before last prompt
     table.insert(messages, #messages, {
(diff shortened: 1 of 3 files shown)


deathbeam:
Added diff display as well, currently with K keybinding that will open float, but there should be a better way to display it for sure:

image

jellydn:
@deathbeam Nice. Let me know when it's ready ;) I will take a try. Thanks.

deathbeam:
> @deathbeam Nice. Let me know when it's ready ;) I will take a try. Thanks.

Just did one last small change I wanted to do to the display so should be ready now I think. I realized that most common user configs will have conflicting K mapping so I changed the keybind to C-d as well and accept to C-y again

PR #103, opened 2024-03-01
feat: Add new healthchecks

deathbeam:
This addresses another issue from MIGRATION.md todo list

- Add curl, and git health checks
- Add dependency health checks
- Add warning for non-nightly neovim as that disables some features

image


jellydn (approved):
LGTM! Thanks @deathbeam

PR #106, opened 2024-03-01
feat: Rework how selectors work so only last buffer is remembered

deathbeam:
This allows updating selection without having to call chat.ask or chat.open again


jellydn (approved):
LGTM! Thanks @deathbeam

PR #110, opened 2024-03-02
feat: Add support for embeddings

### LINKED ISSUE #5: Feature Request: support sub-commands in CopilotChat.nvim
# Context

GitHub Copilot for VS Code offers a range of useful sub-commands like /new, /fix, /explain, and /tests that enhance the coding experience. These commands allow users to generate new code, fix issues, get explanations, and write tests efficiently.

[VSCode](https://gyazo.com/fe85b61f7c78cb1caa299e7024c8b3e9)

# Proposal

Integrate these sub-commands into CopilotChat.nvim to bring similar functionalities to Neovim. Below are the specific commands and their intended functionalities:

- /new: Generate new code blocks or files.
- /fix: Automatically suggest fixes for highlighted code issues.
- /explain: Provide detailed explanations for specific code segments.
- /tests: Generate test cases for existing code.

# Technical Considerations

- We could add a command like `GCFIX`, `GCTests`, etc for each sub command.
- We might need to migrate to pure lua plugin if needed.

# Request for Feedback

I am not sure about the feasibility of this without Github Copilot team input. Let me know if you have any thoughts on plan on this. @gptlang Thanks for great work for bring this plugin to Neovim community. 👏

deathbeam:
Closes #5 

How it works:
1. Grabs outline for all functions for all currently opened buffers
2. Generates embeddings for all the functions
3. Generates embedding from current code selection
4. Uses spatial cosine distance to find similar embeddings to code selection
5. Sends top x matched embeddings as part of the query to copilot in format:

````markdown
query
File: `<filename1>`
```<filetype_for_filename1>
<embeddings_for_filename1>
```
`<filename2>`
```<filetype_for_filename2>
<embeddings_for_filename2>
```
etc
````

TODO:
- [x] Embed all open buffers
- [x] Embed current buffer
- [x] Limit based on token count (wait for tiktoken lua)
- [x] Add config options of what should be embedded

## Example
image

image

## Outline debugging (inside of `CopilotChatDebugInfo`):

image

## Context selection

image

### DIFF
--- a/lua/CopilotChat/context.lua
+++ b/lua/CopilotChat/context.lua
@@ -1,0 +1,223 @@
+local log = require('plenary.log')
+
+local M = {}
+
+local outline_types = {
+  'local_function',
+  'function_item',
+  'arrow_function',
+  'function_definition',
+  'function_declaration',
+  'method_definition',
+  'method_declaration',
+  'constructor_declaration',
+  'class_definition',
+  'class_declaration',
+  'interface_definition',
+  'interface_declaration',
+  'type_alias_declaration',
+  'import_statement',
+  'import_from_statement',
+}
+
+local comment_types = {
+  'comment',
+  'line_comment',
+  'block_comment',
+  'doc_comment',
+}
+
+local ignored_types = {
+  'export_statement',
+}
+
+local off_side_rule_languages = {
+  'python',
+  'coffeescript',
+  'nim',
+  'elm',
+  'curry',
+  'fsharp',
+}
+
+local function spatial_distance_cosine(a, b)
+  local dot_product = 0
+  local magnitude_a = 0
+  local magnitude_b = 0
+  for i = 1, #a do
+    dot_product = dot_product + a[i] * b[i]
+    magnitude_a = magnitude_a + a[i] * a[i]
+    magnitude_b = magnitude_b + b[i] * b[i]
+  end
+  magnitude_a = math.sqrt(magnitude_a)
+  magnitude_b = math.sqrt(magnitude_b)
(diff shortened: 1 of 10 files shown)

### CODE REVIEW

**Comment by deathbeam** [comment]
Added code for generating only outline as well, atm unused just being printed to console

**Comment by deathbeam** [comment]
Ok the draft now uses outline and grabs data for all open buffers

File: lua/CopilotChat/context.lua [hunk 1]
```diff
@@ -0,0 +1,221 @@
...
+local ignored_types = {
+  'export_statement',
+}
+
+local off_side_rule_languages = {
+  'python',
+  'coffeescript',
+  'nim',
+  'elm',
+}
```
gptlang, line 39: Maybe add
- [Miranda](https://en.wikipedia.org/wiki/Miranda_(programming_language)), Haskell's parent
- [Haskell](https://en.wikipedia.org/wiki/Haskell)
- [Curry](https://en.wikipedia.org/wiki/Curry_(programming_language))
- [Elixir](https://en.wikipedia.org/wiki/Elixir_(programming_language)) (, do: blocks)
- [F#](https://en.wikipedia.org/wiki/F_Sharp_(programming_language))
- [Scala](https://en.wikipedia.org/wiki/Scala_(programming_language)) (off-side optional)

deathbeam: This part is only important if function/class do not ends with some ending keyword.So do not applies to haskell and elixir and iassume neither to miranda and as scala is optional probs not there either most of the time. I will add curry and f#

gptlang: I see. Thanks

gptlang (approved):

PR #113, opened 2024-03-02
Tiktoken/Token counting support

gptlang:


lua/CopilotChat/copilot.lua
```diff
@@ -273,12 +278,23 @@ function Copilot:ask(prompt, opts)
...
 end
 
+--- Get the token count for the current selection
+---@param selection string: The selection to count tokens for
+---@param system_prompt string|nil: The system prompt to count tokens for
+function Copilot:get_token_count(selection, system_prompt)
+  if not system_prompt then
+    system_prompt = prompts.COPILOT_INSTRUCTIONS
+  end
+  return self.token_count + Encoder:count(selection) + Encoder:count(system_prompt)
```
deathbeam, line 295: you can just send this in on_done instead on line 231 instead of this being method

lua/CopilotChat/copilot.lua
```diff
@@ -8,6 +8,8 @@ local curl = require('plenary.curl')
 local class = require('CopilotChat.utils').class
 local prompts = require('CopilotChat.prompts')
 
+local Encoder = require('CopilotChat.tiktoken')()
```
deathbeam, line 11: this should be set in construct or it should not be a class at all

lua/CopilotChat/copilot.lua
```diff
@@ -273,12 +278,23 @@ function Copilot:ask(prompt, opts)
       end,
     })
     :after(function()
+      self.token_count = self.token_count + Encoder:count(full_response)
```
deathbeam, line 281: This can also just be moved to before on_done on line 231

lua/CopilotChat/copilot.lua
gptlang: done

lua/CopilotChat/copilot.lua
gptlang: done

lua/CopilotChat/copilot.lua
gptlang: done

deathbeam: I meant to delete this method and send the token count in on_done callback and then just use that in init.lua

lua/CopilotChat/health.lua
```diff
@@ -74,6 +74,11 @@ function M.check()
       'copilot: missing, required for 2 factor authentication. Install copilot.vim or copilot.lua'
     )
   end
+  if lualib_installed('tiktoken_core') then
+    ok('tiktoken_core: installed')
+  else
+    error('tiktoken_core: missing, optional for token counting.')
```
jellydn, line 80:
```suggestion
    warn('tiktoken_core: missing, optional for token counting.')
```

jellydn: It should be `warn` instead of `error`, right? @gptlang

deathbeam (approved):

PR #120, opened 2024-03-05
Move some window-specific logic to chat.lua

deathbeam:
Small cleanup so stuff is in proper place instead of being coupled


lua/CopilotChat/init.lua
```diff
@@ -553,9 +403,73 @@ end
...
+          )
+        end
+      end, { buffer = bufnr })
+    end
+  end)
+
   tiktoken.setup()
   debuginfo.setup()
   M.debug(M.config.debug)
+  M.reset()
```
jellydn, line 472: Hi @deathbeam Do we run reset here?

jellydn, line 472:
```suggestion
```

deathbeam: Yes, it appends help to the buffer, otherwise it would be empty

PR #126, opened 2024-03-06
Remove python plugin from canary as per poll and update migration

deathbeam:
- Removes python plugin as per poll results
- Marks MIGRATION.md authentication as done as per gptlaggg response in Discord


jellydn (approved):
LGTM!

PR #128, opened 2024-03-06
Add back Discord info

deathbeam:
It was part of ChatInPlace note but that one isnt there anymore and badge is nicer


deathbeam:
Hmm the badge doesnt work looks like

PR #168, opened 2024-03-14
Fix broken context extraction for large files.

dmadisetti:
Bug on line 83.

As was, it always returned bool, nil, which would push into the next conditional, which could fall through if the file type is not directly recognized.

In these cases, it's also returning nil. Calling functions treat the response as a table, which leads to unexpected failures.

### DIFF
--- a/lua/CopilotChat/context.lua
+++ b/lua/CopilotChat/context.lua
@@ -80,7 +80,10 @@ function M.build_outline(bufnr)
   local name = vim.api.nvim_buf_get_name(bufnr)
   local ft = vim.bo[bufnr].filetype
   local lang = vim.treesitter.language.get_lang(ft)
-  local ok, parser = lang and pcall(vim.treesitter.get_parser, bufnr, lang) or false, nil
+  local ok, parser = false, nil
+  if lang then
+    ok, parser = pcall(vim.treesitter.get_parser, bufnr, lang)
+  end
   if not ok or not parser then
     ft = string.gsub(ft, 'react', '')
     ok, parser = pcall(vim.treesitter.get_parser, bufnr, ft)
@@ -189,7 +192,7 @@ function M.find_for_query(copilot, opts)
 
   local outline = {}
   if context == 'buffers' then
-    -- For multiiple buffers, only make outlines
+    -- For multiple buffers, only make outlines
     outline = vim.tbl_map(
       function(b)
         return M.build_outline(b)
@@ -208,9 +211,16 @@ function M.find_for_query(copilot, opts)
         filetype = filetype,
       })
     else
-      table.insert(outline, M.build_outline(bufnr))
+      local result = M.build_outline(bufnr)
+      if result ~= nil then
+        table.insert(outline, result)
+      end
     end
   end
+
+  outline = vim.tbl_filter(function(item)
+    return item ~= nil
+  end, outline)
 
   if #outline == 0 then
     on_done({})


lua/CopilotChat/context.lua
```diff
@@ -208,7 +208,8 @@ function M.find_for_query(copilot, opts)
         filetype = filetype,
       })
     else
-      table.insert(outline, M.build_outline(bufnr))
+      -- Since empty just replace with call.
+      outline = M.build_outline(bufnr)
```
deathbeam, line 212: this is incorrect, outline is list of outlines, it do not returns list by itself.

lua/CopilotChat/context.lua
```diff
@@ -154,7 +154,7 @@ function M.build_outline(bufnr)
   get_outline_lines(root)
   local content = table.concat(outline_lines, '\n')
   if content == '' then
-    return
+    return {}
```
deathbeam, line 157: why? Just handle nil outside, do not return empty tables

lua/CopilotChat/context.lua
```diff
@@ -80,12 +80,12 @@ function M.build_outline(bufnr)
...
   local ft = vim.bo[bufnr].filetype
   local lang = vim.treesitter.language.get_lang(ft)
-  local ok, parser = lang and pcall(vim.treesitter.get_parser, bufnr, lang) or false, nil
+  local ok, parser = pcall(vim.treesitter.get_parser, bufnr, lang)
   if not ok or not parser then
     ft = string.gsub(ft, 'react', '')
     ok, parser = pcall(vim.treesitter.get_parser, bufnr, ft)
     if not ok or not parser then
-      return
+      return {}
```
deathbeam, line 88: same here

lua/CopilotChat/context.lua
```diff
@@ -80,12 +80,12 @@ function M.build_outline(bufnr)
   local name = vim.api.nvim_buf_get_name(bufnr)
   local ft = vim.bo[bufnr].filetype
   local lang = vim.treesitter.language.get_lang(ft)
-  local ok, parser = lang and pcall(vim.treesitter.get_parser, bufnr, lang) or false, nil
+  local ok, parser = pcall(vim.treesitter.get_parser, bufnr, lang)
```
deathbeam, line 83: if lang is nil vim.treesitter.get_parser should not be called. if the and/or wasnt working, then just add normal if else around i guess

deathbeam (changes requested):

lua/CopilotChat/context.lua
dmadisetti: from testing, if lang is nil, it's inferred from the buffer. It's non-breaking

but alright

dmadisetti:
Actually tested this time!

lua/CopilotChat/context.lua
```diff
@@ -80,7 +80,10 @@ function M.build_outline(bufnr)
   local name = vim.api.nvim_buf_get_name(bufnr)
   local ft = vim.bo[bufnr].filetype
   local lang = vim.treesitter.language.get_lang(ft)
-  local ok, parser = lang and pcall(vim.treesitter.get_parser, bufnr, lang) or false, nil
+  if not lang then
```
deathbeam, line 83: I did not meant early return but simply to not call the vim.treesitter.get_parser when lang is nil and go straight to fallback. Thats what the original code did (or tried to) e.g avoided the unnecessary call and went to check get_parsr for ft

lua/CopilotChat/context.lua
```diff
@@ -208,11 +211,11 @@ function M.find_for_query(copilot, opts)
         filetype = filetype,
       })
     else
-      table.insert(outline, M.build_outline(bufnr))
+      outline = M.build_outline(bufnr)
```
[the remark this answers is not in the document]
dmadisetti: That's why my original PR sent back an empty list- with your recommendation you have a mixed type signature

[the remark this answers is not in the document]
deathbeam: But you cant sometimes send back empty list and sometimes send back just the outline which isnt a list, thats why the original solution did not worked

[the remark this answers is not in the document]
deathbeam: E.g look at what is build_outline actually returning when it succeeds, I think that is where most of the confusion lies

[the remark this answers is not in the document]
dmadisetti: You're right; optionals seem common in Lua. Added a filter to address `nils`

lua/CopilotChat/context.lua
```diff
@@ -80,7 +80,10 @@ function M.build_outline(bufnr)
   local name = vim.api.nvim_buf_get_name(bufnr)
   local ft = vim.bo[bufnr].filetype
   local lang = vim.treesitter.language.get_lang(ft)
-  local ok, parser = lang and pcall(vim.treesitter.get_parser, bufnr, lang) or false, nil
+  ok, parser = false, nil
```
deathbeam, line 83: This should be local

deathbeam (approved):

jellydn:
@all-contributors add @dmadisetti for code.

PR #171, opened 2024-03-15
fix: avoid nil value happen at posting obj

weykon:
```
    {
        branch = "canary",
        dependencies = {
            { "zbirenbaum/copilot.lua" }, -- or github/copilot.vim
            { "nvim-lua/plenary.nvim" },  -- for curl, log wrapper
        },
        opts = {
            debug = true, -- Enable debugging
            -- See Configuration section for rest
        },
        config = function()
            require("CopilotChat").setup({ debug = true })
            vim.keymap.set("v", "<leader>ai", "<cmd>CopilotChatToggle<CR>")
        end
    },
```
I was using lunarvim to load it into use, and the configuration is as above. 
Every time I enter prompt after selection and then press enter, there will be errors.
I looked it up and tried to solve it like this. 
I am not familiar with lua, so I am not sure if it is reasonable to modify it like this, but it has no errors.
Please give me some advice on how to discover the problems and modify them.
Thanks !

### DIFF
--- a/lua/CopilotChat/copilot.lua
+++ b/lua/CopilotChat/copilot.lua
@@ -445,9 +445,13 @@ function Copilot:ask(prompt, opts)
         full_response = full_response .. content
       end,
     })
-    :after(function()
-      self.current_job = nil
-    end)
+    if self.current_job then
+      self.current_job:after(
+        function ()
+          self.current_job = nil
+        end
+      )
+    end
 end
 
 --- Generate embeddings for the given inputs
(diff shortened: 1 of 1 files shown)


deathbeam:
I dont have this issue, is it possible that your plenary.nvim dependency is outdated?

weykon:
It may be this problem. I saw that the commit of plenary.nvim is 2023, which is relatively old, but when I reinstalled it, he still installed that version. I still don't know how to set them. 
I had update Lvim and Plugins with update command

deathbeam:
yea, the plugin requires very recent plenary version to function properly. I dont use lazy.nvim so not sure what could be the problem and why is it not updating but maybe @jellydn knows?

jellydn:
You could use `commit=f7adfc4b3f4f91aab6caebf42b3682945fbc35be` with lazy.nvim to pin to the latest commit @weykon

weykon:
I switched to f7adfc4 you said, the function can be used normally. thank you both!

PR #188, opened 2024-03-18
 Add relativenumber to config

### LINKED ISSUE #187: Allow setting relativenumber in config
The chat window has relativenumber off by default. Allow changing this option in the config. 

This would make it more comfortable for us vim users that use relative numbers to move around. I find it particularly useful when yanking just a few lines from the suggestion in the chat window. 

Thanks for creating this plugin!

tlacuilose:
Adds a window options field to the config that allows setting relative numbers for the chat window. 

```lua
  -- default window options
  window = {
    layout = 'vertical', -- 'vertical', 'horizontal', 'float'
    -- Options below only apply to floating windows
    relative = 'editor', -- 'editor', 'win', 'cursor', 'mouse'
    border = 'single', -- 'none', single', 'double', 'rounded', 'solid', 'shadow'
    width = 0.8, -- fractional width of parent
    height = 0.6, -- fractional height of parent
    row = nil, -- row position of the window, default is centered
    col = nil, -- column position of the window, default is centered
    title = 'Copilot Chat', -- title of chat window
    footer = nil, -- footer of chat window
    zindex = 1, -- determines if window is on top or below other floating windows
    options = {
      relativenumber = false, -- show relative line numbers
    },
  },
```
Resolves #187 
### How it looks
Screenshot 2024-03-17 at 11 04 55 p m

### DIFF
--- a/README.md
+++ b/README.md
@@ -240,6 +240,9 @@ Also see [here](/lua/CopilotChat/config.lua):
     title = 'Copilot Chat', -- title of chat window
     footer = nil, -- footer of chat window
     zindex = 1, -- determines if window is on top or below other floating windows
+    options = {
+      relativenumber = false, -- show relative line numbers
+    },
   },
   -- default mappings
   mappings = {
--- a/lua/CopilotChat/chat.lua
+++ b/lua/CopilotChat/chat.lua
@@ -152,12 +152,14 @@ function Chat:open(config)
     self.winnr = vim.api.nvim_open_win(self.bufnr, false, win_opts)
   end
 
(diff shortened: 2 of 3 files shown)


jellydn:
Thanks, @tlacuilose Could you move the other buffer options to config as well? e.g.: `wrap`, `linebreak`, etc.

deathbeam:
Btw cant you simplu use autocmd BufEnter on copilot-chat? Thats how you work with buffers normally after all

tlacuilose:
TIL 😅. I will use that, thanks for the suggestion!

PR #195, opened 2024-03-18
194 add callback response

### LINKED ISSUE #194: Feature request: add callback function to ask API
Would be great if there was an optional callback function that would receive the response.
This would enable users to automate things with the response, e.x. take the commit message and use it in `git commit -m <response>`. Perhaps something simple like:
```
            require("CopilotChat").ask(input, {
               selection = ..,
               callback = function(response) {
                   vim.cmd("!git commit -m " .. response)
                }
            })
```

aweis89:
Fixes #194

### DIFF
--- a/README.md
+++ b/README.md
@@ -118,7 +118,7 @@ See @deathbeam for [configuration](https://github.com/deathbeam/dotfiles/blob/ma
 
 ### API
 
-````lua
+```lua
 local chat = require("CopilotChat")
 
 -- Open chat window
@@ -158,16 +158,9 @@ chat.ask("Explain how it works.", {
 })
 
 -- Ask a question and do something with the response
-local prompt = require("CopilotChat.config").prompts.Commit.prompt
-chat.ask(prompt, {
-  selection = require("CopilotChat.select").gitdiff,
+chat.ask("Show me something interesting", {
   callback = function(response)
-    local gitcommit = response:match("```gitcommit\n(.*)```")
-    if gitcommit then
-      vim.api.nvim_command("Git commit -m " .. '"' .. gitcommit .. '"')
-    else
-      print("No commit message found.")
-    end
+    print("Response:", response)
   end,
 })
 
@@ -182,7 +175,7 @@ actions.pick(actions.help_actions())
 
 -- Pick prompt actions
 actions.pick(actions.prompt_actions())
-````
+```
 
 ## Configuration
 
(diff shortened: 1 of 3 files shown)


lua/CopilotChat/config.lua
```diff
@@ -63,6 +63,8 @@ local select = require('CopilotChat.select')
 ---@field selection nil|fun(source: CopilotChat.config.source):CopilotChat.config.selection?
 ---@field window CopilotChat.config.window?
 ---@field mappings CopilotChat.config.mappings?
+---@field callback fun(response: string)?
+
```
deathbeam, line 67: Remove the space

lua/CopilotChat/config.lua
```diff
@@ -63,6 +63,8 @@ local select = require('CopilotChat.select')
 ---@field selection nil|fun(source: CopilotChat.config.source):CopilotChat.config.selection?
 ---@field window CopilotChat.config.window?
 ---@field mappings CopilotChat.config.mappings?
+---@field callback fun(response: string)?
```
deathbeam, line 66: Probs should define also default value in config as it is part of config spec (which includes also updating README). Also probs put it under selection so window and mappings stay at very bottom

deathbeam (changes requested):
On top of this I think just small example in API README section would be nice (similar to the existing .ask examples there)

README.md
```diff
@@ -168,7 +182,7 @@ actions.pick(actions.help_actions())
 
 -- Pick prompt actions
 actions.pick(actions.prompt_actions())
-```
+````
```
aweis89, line 185:
```suggestion
```
```

File: lua/CopilotChat/config.lua (hunk 2 above)
Comment (aweis89) [reply lua/CopilotChat/config.lua]: Not sure I follow, you mean an empty function that does nothing?
Is the optional nil `?` and handling that scenario not sufficient?

Comment (deathbeam) [reply lua/CopilotChat/config.lua]: Just callback = nil as default value yea, for ppl without LSP who are just looking at the config directly (and also for the README part where it shows just config)

File: README.md (hunk 3 above)
Comment (deathbeam) [inline README.md:185]: This should be just trple backtick no?

Comment (deathbeam) [reply README.md:185]: Oh nvm I see

File: README.md [hunk 4]
```diff
@@ -157,6 +157,20 @@ chat.ask("Explain how it works.", {
   selection = require("CopilotChat.select").buffer,
 })
 
+-- Ask a question and do something with the response
+local prompt = require("CopilotChat.config").prompts.Commit.prompt
```
deathbeam, line 161: This example can be just super simple, the git stuff can just be tip later maybe but I would simply just do `vim.print(response)` as example here, e.g super simple, without selection too

deathbeam:
Adjusted the PR a bit with the changes i proposed

deathbeam (approved):

deathbeam:
Hmm, let me try this I guess

@all-contributors add @aweis89 for code and doc.

PR #201, opened 2024-03-19
docs: Make configuration easier to navigate through

deathbeam:
Split configuration to section with newlines
Group similar configs together in sections


jellydn (approved):
LGTM!

PR #202, opened 2024-03-20
feat: Show token count as virtual line instead of in buffer

deathbeam:
No reason to write out the tokens to buffer,and it also messes up the history when saving chats so just show it as virtual line instead

image


jellydn (approved):
Nice. Thanks @deathbeam

PR #204, opened 2024-03-20
203 add custom mappings

### LINKED ISSUE #203: Feature request: support custom mappings
It would be awesome if we could add custom mappings for the copilot buffer which allows users to map key combos to custom functions that get passed the previous response.
Some common use-cases would be to copy the previous code block, or do something custom with the prior response.
Perhaps a new field in the config (or maybe use the existing `mappings` table, wdyt?):

```lua
user_mappings = {
  -- copy to + register
  ["<C-g>"] = function(response)
     local to_copy =  res:match("```.*\n(.-)```")
     vim.fn.setreg("+",  to_copy)
  end
}
```

Thanks again for creating one of my all-time favorite plugins I use daily!

aweis89:
Closes #203

### DIFF
--- a/README.md
+++ b/README.md
@@ -335,8 +335,11 @@ vim.api.nvim_create_autocmd('BufEnter', {
     pattern = 'copilot-*',
     callback = function()
         vim.opt_local.relativenumber = true
-        -- or do something with the last response
-        print(require("CopilotChat").get_last_response())
+
+        -- C-p to print last response
+        vim.keymap.set('n', '<C-p>', function()
+          print(require("CopilotChat").response())
+        end, { buffer = true, remap = true })
     end
 })
 ```
--- a/README.md
+++ b/README.md
@@ -341,7 +341,8 @@ vim.api.nvim_create_autocmd('BufEnter', {
 ```
 
 ### Custom mappings
-```lua
+
+````lua
 require("CopilotChat").setup({
   user_mappings = {
     ["<C-w>"] = function(response)
@@ -354,7 +355,7 @@ require("CopilotChat").setup({
     end,
   }
 })
-```
+````
 
 ## Tips
 
--- a/README.md
+++ b/README.md
@@ -203,6 +203,7 @@ Also see [here](/lua/CopilotChat/config.lua):
   context = nil, -- Default context to use, 'buffers', 'buffer' or none (can be specified manually in prompt via @).
   history_path = vim.fn.stdpath('data') .. '/copilotchat_history', -- Default path to stored history
   callback = nil, -- Callback to use when ask response is received
+  user_mappings = {}, -- User mappings to use
 
   -- default selection (visual or line)
   selection = function(source)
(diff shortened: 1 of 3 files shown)


lua/CopilotChat/init.lua
```diff
@@ -349,6 +349,15 @@ function M.ask(prompt, config, source)
             if config.callback then
               config.callback(response)
             end
+            for key, value in pairs(config.user_mappings) do
```
deathbeam, line 352: This creates bunch of mappings every time you get a response which I highly doubt is the correct approach. I think better solution would be to expose getting the last response (look at how its done in the existing mapping) and then just let ppl use that and make their own keymaps similar to custom buffer options on BufEnter etc.

lua/CopilotChat/init.lua
```diff
@@ -671,6 +676,21 @@ function M.setup(config)
   debuginfo.setup()
   M.debug(M.config.debug)
 
+  for mapping, val in pairs(M.config.user_mappings) do
+    vim.api.nvim_create_autocmd('BufEnter', {
```
aweis89, line 680: Not sure if this is the intended pattern, but I noticed when I just use `vim.keymap.set`, it doesn't override global pre-existing keymaps. I _think_ this may be the case for other mappings as well (which seems undesirable to me, wdyt?)

deathbeam: Well vim.keymap.set uses remap instead of noremap so that was probably your issue. But also I meant that last response could just be exposed from the API so users can set the custom mappings from outside of the plugin.

README.md
```diff
@@ -335,6 +335,8 @@ vim.api.nvim_create_autocmd('BufEnter', {
     pattern = 'copilot-*',
     callback = function()
         vim.opt_local.relativenumber = true
+        -- or do something with the last response
+        print(require("CopilotChat").get_last_response())
```
deathbeam, line 339: this could just be full example of the keymap I think, so something like

```lua
-- C-p to print last response
vim.keymap.set('n', '<C-p>', function()
  print(require("CopilotChat").get_last_response())
end, { buffer = true, remap = true })
```

lua/CopilotChat/init.lua
```diff
@@ -268,6 +272,11 @@ function M.toggle(config, source)
   end
 end
 
+-- @returns string
+function M.get_last_response()
```
deathbeam, line 276: This should probs follow the `prompts` naming and should just be `M.response()` imo

lua/CopilotChat/init.lua
```diff
@@ -31,6 +32,9 @@ local state = {
   last_system_prompt = nil,
   last_code_output = nil,
 
+  -- Response for mappings
+  response = nil,
```
deathbeam, line 36: You should also probs reset this in M.reset()

deathbeam (changes requested):

deathbeam (approved):

PR #210, opened 2024-03-20
fix: Use prompt selection when picking prompts

### LINKED ISSUE #206: Telescope prompt_actions not use select.gitdiff selection
```lua
local actions = require "CopilotChat.actions"
require("CopilotChat.integrations.telescope").pick(actions.prompt_actions())
```

When i run above code, it show list prompt action, but when choose `Commit` or `CommitStaged`, it open chat window window but with `visual` selection, not `gitdiff`

It work fine if I use command `:CopilotChatCommit`

deathbeam:
Instead of definining the selection when generating actions and hardcoding it, use prompt configuration for grabbing selection.

Closes #206


jellydn (approved):
LGTM!

PR #213, opened 2024-03-21
fix: Add back option to configure selection for picker actions

deathbeam:


jellydn (approved):
Thanks @deathbeam for quick action :)

PR #217, opened 2024-03-21
feat: Add start and end line number as line range to active selection

### LINKED ISSUE #214: Feature request add line numbers to buffer selection
I noticed when prompting, using buffer as the selection, copilot often mixes up the correct line numbers when it references it in the prompt or response.
It seems like when adding line numbers to the selection it's no longer an issue.
Given that referencing lines may be a common use-case, perhaps we should change the default?

deathbeam:
Closes #214
Closes #215


jellydn:
Hi @deathbeam is this ready?

deathbeam:
It is yea

PR #223, opened 2024-03-22
fix: Properly start insert mode on opening window, regardless of source

### LINKED ISSUE #222: `auto_insert_mode` does not start insert mode when opening window
Setting `auto_insert_mode` to `true` does not seem to start insert mode when opening a new window using `:CopilotChat`. I have tried debugging the issue myself, and it seems like `M.ask()` always calls `M.open()` with `no_focus = true`, and there is an `if not no_focus` check which prevents `startinsert` from being called. I'm not sure if this is intended or not, but the documentation for `auto_insert_mode` suggests otherwise.

deathbeam:
Closes #222


jellydn (approved):
Thanks @deathbeam

PR #240, opened 2024-03-26
feat: toggleable "done!" notification

D7ry:
I personally found the "done!" notification a bit distractive, so introduced a new option `notify_done` that makes this feature toggleable. It's on by default.


D7ry:
oops i think there's a little bug in my impl; closing it for now

PR #241, opened 2024-03-26
Remove "done!" notification

D7ry:
By configuring `notify_done` variable in opts, users can disable the "done!" notification that some may find a bit distracting.

### DIFF
--- a/README.md
+++ b/README.md
@@ -203,7 +203,6 @@ Also see [here](/lua/CopilotChat/config.lua):
   error_header = '**Error** ', -- Header to use for errors
   separator = '---', -- Separator to use in chat
 
-  notify_done = true, -- Send notification when chat is complete
 
   show_folds = true, -- Shows folds for sections in chat
   show_help = true, -- Shows help message as virtual lines when waiting for user input
--- a/README.md
+++ b/README.md
@@ -203,6 +203,8 @@ Also see [here](/lua/CopilotChat/config.lua):
   error_header = '**Error** ', -- Header to use for errors
   separator = '---', -- Separator to use in chat
 
+  notify_done = true, -- Send notification when chat is complete
+
   show_folds = true, -- Shows folds for sections in chat
   show_help = true, -- Shows help message as virtual lines when waiting for user input
   auto_follow_cursor = true, -- Auto-follow cursor in chat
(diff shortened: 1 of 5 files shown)


deathbeam:
I think way better way would be to remove the notify completely from where its being called now and just set default callback (thats currently by default nil) to be function that calls `vim.notify`.

deathbeam (changes requested):
As i said in my response, should just be callback imo

D7ry:
> As i said in my response, should just be callback imo

I believe just not setting the default callback is better; removed the notify completely!

deathbeam:
Yea i agree, much cleaner solution

deathbeam (approved):

lua/CopilotChat/config.lua
```diff
@@ -76,6 +76,7 @@ local select = require('CopilotChat.select')
 ---@field prompts table<string, CopilotChat.config.prompt|string>?
 ---@field window CopilotChat.config.window?
 ---@field mappings CopilotChat.config.mappings?
+
```
deathbeam, line 79: Same here

lua/CopilotChat/chat.lua
```diff
@@ -30,7 +30,7 @@ function CopilotChatFoldExpr(lnum, separator)
   return '='
 end
 
-local Chat = class(function(self, mark_ns, help, on_buf_create)
+local Chat = class(function(self, mark_ns, help, on_buf_create, config)
```
deathbeam, line 33: Im not sure why are you passing config here

D7ry: updated. the config was required previous for getting the option but no longer

deathbeam (changes requested):
see comments

README.md
```diff
@@ -203,6 +203,7 @@ Also see [here](/lua/CopilotChat/config.lua):
   error_header = '**Error** ', -- Header to use for errors
   separator = '---', -- Separator to use in chat
 
+
```
jellydn, line 206:
```suggestion
```

lua/CopilotChat/config.lua
jellydn, line 79:
```suggestion
```

jellydn:
@all-contributors add @D7ry for code.

PR #247, opened 2024-03-28
fix: Remap bindings to not conflict with terminal

### LINKED ISSUE #246: <CR> in insert mode submits the prompt instead of inserting a new line
When I hit `<CR>` in insert mode in the chat window, it submits.

The only other binding I have on `<CR>` for imap is nvim-cmp.

Peek 2024-03-28 13-44

deathbeam:
C-m is same as CR on a lot of terminals, remap to C-s

Closes #246

### DIFF
--- a/lua/CopilotChat/config.lua
+++ b/lua/CopilotChat/config.lua
@@ -169,7 +169,7 @@ return {
     },
     submit_prompt = {
       normal = '<CR>',
-      insert = '<C-m>',
+      insert = '<C-s>',
     },
     accept_diff = {
       normal = '<C-y>',


lua/CopilotChat/config.lua
```diff
@@ -169,7 +169,7 @@ return {
     },
     submit_prompt = {
       normal = '<CR>',
-      insert = '<C-m>',
+      insert = '<C-s>',
```
jellydn, line 172: Oh, I didn't get this error as my config is `'<C-Enter>'`.

PR #252, opened 2024-03-30
Fix telescope preview if prompt contains newlines

ornew:
If the prompt contains newlines, the error
`'replacement string' item contains newlines` will occur when previewing the Telescope integration. This is because `nvim_buf_set_lines` prohibits newlines from being included in the content it sets. Instead, this function takes a list of strings as arguments.

In this commit, vim.split splits strings with newlines. With this fix, Telescope preview will work correctly even when the prompt contains newlines.


deathbeam (approved):

jellydn:
Thanks @ornew.

@all-contributors add @ornew for code.

PR #266, opened 2024-04-09
feat: Send whole content of small files in @buffers as well

deathbeam:
Uses same logic as @buffer used for smaller files. This also reduces the small file threshold, 1k lines was probs a bit too much.


jellydn (approved):
LGTM!

PR #270, opened 2024-04-10
docs: Match README config with config.lua

deathbeam:
For chat headers, config.lua was updated before but README wasnt


jellydn (approved):
LGTM!

PR #276, opened 2024-04-12
feat(copilot): load GitHub token from environment

lingjie00:
The copilot module now first tries to load the GitHub token from the environment variable 'GITHUB_TOKEN'. If the variable is not set, it falls back to loading the token from the file.

### DIFF
--- a/lua/CopilotChat/copilot.lua
+++ b/lua/CopilotChat/copilot.lua
@@ -91,6 +91,13 @@ local function find_config_path()
 end
 
 local function get_cached_token()
+-- loading token from the environment
+  local token = os.getenv('GITHUB_TOKEN')
+  if token then
+    return token
+  end
+
+-- loading token from the file
   local config_path = find_config_path()
   if not config_path then
     return nil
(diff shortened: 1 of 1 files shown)


gptlang:
Can we name it something other than GITHUB_TOKEN? That's a common environment variable name used by other programs.

lingjie00:
> Can we name it something other than GITHUB_TOKEN? That's a common environment variable name used by other programs.

Initially I was thinking to re-use the GITHUB_TOKEN already in the environment. Do you think we should let user to maybe choose the variable in config while setting the default as GITHUB_TOKEN?

deathbeam:
It definitely should not be configurable, we should just follow whatever github does. Which means we should also check for the CODESPACE env variable like the copilot.vim does.

deathbeam (approved):

acheong08 (approved):
For reference:

```js
                async getGitHubToken(t) {
                        if (process.env.CODESPACES === "true" && process.env.GITHUB_TOKEN)
                                return { token: process.env.GITHUB_TOKEN };
```
in https://github.com/github/copilot.vim/blob/release/dist/agent.js

Edit: Oops wrong account

jellydn:
@all-contributors add @lingjie00 for code.

PR #278, opened 2024-04-12
allow height and width for horizontal and vertical

devshoe:
This should allow for setting height and width from config for vertical and horizontal layouts also, so that it has a smaller width. Width only applicable to vertical and height for horizontal. I think it just makes some more sense.

### DIFF
--- a/lua/CopilotChat/chat.lua
+++ b/lua/CopilotChat/chat.lua
@@ -134,7 +134,12 @@ function Chat:open(config)
   elseif layout == 'vertical' then
     if is_stable() then
       local orig = vim.api.nvim_get_current_win()
-      vim.cmd('vsplit')
+      local cmd = 'vsplit'
+      if config.window.width ~= nil and config.window.width ~= 0 then
+        local width = math.floor(vim.o.columns * config.window.width)
+        cmd = string.format('%dvsplit', width)
+      end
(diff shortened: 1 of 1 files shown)


lua/CopilotChat/chat.lua
```diff
@@ -134,7 +134,12 @@ function Chat:open(config)
   elseif layout == 'vertical' then
     if is_stable() then
       local orig = vim.api.nvim_get_current_win()
-      vim.cmd('vsplit')
+      local cmd = 'vsplit'
+      if config.window.width ~= nil and config.window.width ~= 0 then
```
deathbeam, line 138: You only added support for stable, needs to be done for neovim nightly as well, e.g the else block

lua/CopilotChat/chat.lua
```diff
@@ -144,7 +149,12 @@ function Chat:open(config)
   elseif layout == 'horizontal' then
     if is_stable() then
       local orig = vim.api.nvim_get_current_win()
-      vim.cmd('split')
+      local cmd = 'split'
```
deathbeam, line 152: Same here

deathbeam (changes requested):
Also add comment to config.lua and README.md section with config.lua with new width/height behaviour so its documented

PR #304, opened 2024-05-06
feat: Add support for select.clipboard

deathbeam:
Useful for providing system clipboard contents easily (for example for using CopilotChat quickly in replace mode in tmux popup etc)


jellydn (approved):
LGTM!

PR #308, opened 2024-05-08
feat: Add method and command to stop current copilot output

deathbeam:
See https://github.com/CopilotC-Nvim/CopilotChat.nvim/discussions/307


jellydn (approved):
LGTM!

PR #322, opened 2024-05-17
fix: consider 0.10 as stable and 0.11 as nightly

### LINKED ISSUE #321: Failed a `checkhealth` operation
image

I just updated to neovim 0.10 (stable) and I encounter this issue. I tried calling `:Lazy sync` and reinstalling a plugin, but it didn't help.

Seems like it can't regex match vim version, because it doesn't consider 0.10 as stable:
https://github.com/CopilotC-Nvim/CopilotChat.nvim/blob/af18fa2fbc765ddf9bacd22e15c974642f20d19b/lua/CopilotChat/health.lua#L48-L49

frolvanya:
closes #321 

Let's consider 0.10 as stable version from now on. I suppose 0.9.5 still can be considered as stable for a few months until majority of people will migrate to 0.10

### DIFF
--- a/lua/CopilotChat/health.lua
+++ b/lua/CopilotChat/health.lua
@@ -41,11 +41,12 @@ end
 function M.check()
   start('CopilotChat.nvim [core]')
 
-  local is_nightly = vim.fn.has('nvim-0.11') == 1
-  local is_good_stable = vim.fn.has('nvim-0.9.5') == 1 or vim.fn.has('nvim-0.10') == 1
   local vim_version = vim.api.nvim_command_output('version')
-  if is_nightly then
-    local dev_number = tonumber(vim_version:match('dev%-(%d+)'))
+  local is_good_stable = vim.fn.has('nvim-0.9.5') == 1 or vim.fn.has('nvim-0.10.0') == 1
+
+  local dev_number = tonumber(vim_version:match('dev%-(%d+)'))
+
+  if dev_number then
     if dev_number >= 2500 then
       ok('nvim: ' .. vim_version)
     else
--- a/lua/CopilotChat/health.lua
+++ b/lua/CopilotChat/health.lua
@@ -41,8 +41,8 @@ end
 function M.check()
   start('CopilotChat.nvim [core]')
 
-  local is_nightly = vim.fn.has('nvim-0.10.0') == 1
-  local is_good_stable = vim.fn.has('nvim-0.9.5') == 1
+  local is_nightly = vim.fn.has('nvim-0.11') == 1
+  local is_good_stable = vim.fn.has('nvim-0.9.5') == 1 or vim.fn.has('nvim-0.10') == 1
   local vim_version = vim.api.nvim_command_output('version')
   if is_nightly then
     local dev_number = tonumber(vim_version:match('dev%-(%d+)'))


lua/CopilotChat/health.lua
```diff
@@ -41,8 +41,8 @@ end
 function M.check()
   start('CopilotChat.nvim [core]')
 
-  local is_nightly = vim.fn.has('nvim-0.10.0') == 1
-  local is_good_stable = vim.fn.has('nvim-0.9.5') == 1
+  local is_nightly = vim.fn.has('nvim-0.11') == 1
+  local is_good_stable = vim.fn.has('nvim-0.9.5') == 1 or vim.fn.has('nvim-0.10') == 1
```
deathbeam, line 45: is this necessary? 0.10 will already have 0.9.5 no?

lua/CopilotChat/health.lua
```diff
@@ -41,8 +41,8 @@ end
 function M.check()
   start('CopilotChat.nvim [core]')
 
-  local is_nightly = vim.fn.has('nvim-0.10.0') == 1
-  local is_good_stable = vim.fn.has('nvim-0.9.5') == 1
+  local is_nightly = vim.fn.has('nvim-0.11') == 1
```
deathbeam, line 44: Thinking about this, maybe is_nightly check should be simply vim_version:match('dev%-(%d+)'). If its not nil then at the dev_number = just do tonumber(is_nightly) etc so we dont have to change the current nightly version all the time

deathbeam (changes requested):

lua/CopilotChat/health.lua
frolvanya: Done
https://github.com/CopilotC-Nvim/CopilotChat.nvim/blob/becc76f54cab703e031a6978f8fd639280ae6db0/lua/CopilotChat/health.lua#L44-L61

lua/CopilotChat/health.lua
frolvanya: It will not find dev tag in 0.10 version as it's stable now and it won't consider it as stable, therefore we need to check if the version is either 0.9.5 or 0.10.0

lua/CopilotChat/health.lua
```diff
@@ -41,11 +41,12 @@ end
...
   start('CopilotChat.nvim [core]')
 
-  local is_nightly = vim.fn.has('nvim-0.10.0') == 1
-  local is_good_stable = vim.fn.has('nvim-0.9.5') == 1
   local vim_version = vim.api.nvim_command_output('version')
-  if is_nightly then
-    local dev_number = tonumber(vim_version:match('dev%-(%d+)'))
+  local is_good_stable = vim.fn.has('nvim-0.9.5') == 1 or vim.fn.has('nvim-0.10.0') == 1
+
+  local dev_number = tonumber(vim_version:match('dev%-(%d+)'))
```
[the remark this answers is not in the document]
frolvanya: `tonumber` return optional number, so it's safe to do it. Basically tonumber(nil) == nil, so `if dev_number` will check if dev_number exists and it's possible to parse it as int

image
image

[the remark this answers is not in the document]
frolvanya: But if you want, I can to this. IMO, it just increases code length

```lua
local dev_version_str = vim_version:match('dev%-(%d+)')

if dev_version_str then
  local dev_number = tonumber(dev_version_str)
  if dev_number and dev_number >= 2500 then
    ok('nvim: ' .. vim_version)
  else
    error(
      'nvim: outdated, please upgrade to an up-to-date nightly version. See "https://github.com/neovim/neovim".'
    )
  end
elseif is_good_stable then
  ok('nvim: ' .. vim_version)
else
  error('nvim: unsupported, please upgrade to 0.9.5 or later. See "https://neovim.io/".')
end
```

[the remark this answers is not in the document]
deathbeam: Ah alright, I thought tonumber will fail on nil. Looks good then

WillEhrendreich:
I'm on a nightly of 0.11.0 but checkhealth fails and says I should upgrade. 
image

deathbeam:
Oh the commit number is since last release, need to fix that

frolvanya:
> Oh the commit number is since last release, need to fix that

Oops, I messed up. Do you want me to fix it? If so, do you want to completely remove this section with `>= 2500` or replace it with `>= 67` (since neovim-0.11 has dev number 67)

PR #332, opened 2024-05-27
feat: highlight the selection in the source buffer

folke:
I added a new option `highlight_selection`, that highlights the selection in the source buffer when in the chat window.
The highlight is only visible when the chat window is focused.

The hl group for the selection is `CopilotChatSelection` and is linked to `Visual` by default.

The option is enabled by default.

Let me know what you think!

image

### DIFF
--- a/lua/CopilotChat/init.lua
+++ b/lua/CopilotChat/init.lua
@@ -275,6 +275,34 @@ function M.prompts(skip_system)
   return prompts_to_use
 end
 
+local selection_ns = nil ---@type number
+
+--- Highlights the selection in the source buffer.
+---@param clear? boolean
+function M.highlight_selection(clear)
+  for _, buf in ipairs(vim.api.nvim_list_bufs()) do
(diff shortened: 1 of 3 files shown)


lua/CopilotChat/init.lua
```diff
@@ -275,6 +275,34 @@ function M.prompts(skip_system)
   return prompts_to_use
 end
 
+local selection_ns = vim.api.nvim_create_namespace('copilot-chat-selection')
+
+--- Highlights the selection in the source buffer.
+---@param opts? {clear?: boolean}
+function M.highlight_selection(opts)
```
deathbeam, line 282: Opts can just be boolean, if highlight will be more configurable in future we usually put defaults to config and then allow config object to be passed around

deathbeam (changes requested):
Otherwise looks good

folke:
All done!

deathbeam (approved):
That was quick, looks good thanks!

deathbeam:
@all-contributors add @folke for code, docs

folke:
Thanks for merging!

PR #336, opened 2024-05-27
feat: enhanced header rendering in the chat window

### LINKED ISSUE #334: feature: enhanced rendering of the chat widow
Are you interested in a PR that allows customizing the chat window like in the screenshot?

image

folke:
As discussed, the separator is now repeated in a virtual line and has hl group `CopilotChatSeparator`.

The headers have `hl_group` `CopilotChatHeader`.

I didn't change the default header strings, so they still have `##`, but that's no longer needed to get the same styling.
 
image

With custom header strings:

image

Closes #334

### DIFF
--- a/lua/CopilotChat/chat.lua
+++ b/lua/CopilotChat/chat.lua
@@ -1,6 +1,7 @@
 ---@class CopilotChat.Chat
 ---@field bufnr number
 ---@field winnr number
+---@field separator string
 ---@field spinner CopilotChat.Spinner
 ---@field valid fun(self: CopilotChat.Chat)
 ---@field visible fun(self: CopilotChat.Chat)
@@ -33,11 +34,13 @@ end
 
 local Chat = class(function(self, mark_ns, help, on_buf_create)
   self.mark_ns = mark_ns
+  self.header_ns = vim.api.nvim_create_namespace('copilot-chat-headers')
   self.help = help
   self.on_buf_create = on_buf_create
   self.bufnr = nil
   self.winnr = nil
   self.spinner = nil
+  self.separator = nil
 
   self.buf_create = function()
     local bufnr = vim.api.nvim_create_buf(false, true)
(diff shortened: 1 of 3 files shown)


lua/CopilotChat/chat.lua
```diff
@@ -33,11 +34,13 @@ end
 
 local Chat = class(function(self, mark_ns, help, on_buf_create)
   self.mark_ns = mark_ns
+  self.header_ns = vim.api.nvim_create_namespace('copilot-chat-headers')
```
folke, line 37: Creating a namespace returns the existing one when it already exists, so no need to create it during setup and then pass it down.

lua/CopilotChat/chat.lua
```diff
@@ -66,6 +69,33 @@ function Chat:visible()
     and vim.api.nvim_win_get_buf(self.winnr) == self.bufnr
 end
 
+function Chat:render()
+  if not self:visible() then
+    return
+  end
+  vim.api.nvim_buf_clear_namespace(self.bufnr, self.header_ns, 0, -1)
+  local lines = vim.api.nvim_buf_get_lines(self.bufnr, 0, -1, false)
```
deathbeam, line 77: Hmm would it be better to add the extmark when the headers are being appended to the chat instead of rereading the whole buffer every time? If i understand extmarks correctly they should persist just fine so readding them after every text change should not be necessary

deathbeam (changes requested):

lua/CopilotChat/chat.lua
folke: I'd keep it like this, since this is a really fast operation and will ensure the extmarks are always correct.

Plugins like lazy.nvim, trouble.nvim and other I didnt write, all basically re-render the whole buffer whenever something in their data changes. For trouble and lazy this means a lot of extmarks. The extmarks for copilot chat are nothing compared to that :)

folke: Same with treesitter. Simple changes in your source file, can mean lots of updates in terms of extmarks.

deathbeam: Interesting, I assumed it can get pretty bad with a lot of lines and extmarks but if its quick then sounds good as this way its a lot simpler to manage.

deathbeam (approved):

GitMurf:
@deathbeam will this effect saving a chat conversation? Historically you could save the buffer as markdown file and have a nicely formatted conversation in markdown. Is that no longer the case if we are removing the ## Heading Syntax?

deathbeam:
> @deathbeam will this effect saving a chat conversation? Historically you could save the buffer as markdown file and have a nicely formatted conversation in markdown. Is that no longer the case if we are removing the ## Heading Syntax?

The syntax is still there this just added some extmarks on top for better visuals and thats it, do not affects how the data are stored at all

GitMurf:
> The syntax is still there this just added some extmarks on top for better visuals and thats it, do not affects how the data are stored at all

Oh ok. I guess I got a little confused with folke's custom header strings screenshot above.

PR #343, opened 2024-05-28
fix(telescope): ensure correct selection with telescope prompt actions

### LINKED ISSUE #329: Code doesn't get passed if using the the prompt_actions when using the Telescope integration
### Discussed in https://github.com/CopilotC-Nvim/CopilotChat.nvim/discussions/288

<div type='discussions-op-text'>

<sup>Originally posted by **Andrej-Marsic** April 19, 2024</sup>
I assume in some cases it's the visual selection that gets passed to the prompt, but if one tries to use it through the telescope integration, the selection disappears and nothing gets passed (or the current line at most)</div>

folke:
Fixes #329

### DIFF
--- a/lua/CopilotChat/integrations/telescope.lua
+++ b/lua/CopilotChat/integrations/telescope.lua
@@ -17,6 +17,10 @@ function M.pick(pick_actions, opts)
     return
   end
 
+  if vim.fn.mode():lower():find('v') then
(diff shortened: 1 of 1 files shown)


deathbeam:
Thanks. I guess this is slightly better than feeding <esc> like we do on other places

PR #344, opened 2024-05-28
chore: Use same way to exit visual mode everywhere and move to utils

deathbeam:


deathbeam:
hmm actually it looks like this does not work when i have chat already open and use .ask from keymap

PR #345, opened 2024-05-28
feat: Add nvim-cmp integration

deathbeam:
image

### DIFF
--- a/README.md
+++ b/README.md
@@ -501,6 +501,30 @@ Requires [fzf-lua](https://github.com/ibhagwan/fzf-lua) plugin to be installed.
 
 </details>
 
+<details>
+<summary>nvim-cmp integration</summary>
+
+Requires [nvim-cmp](https://github.com/hrsh7th/nvim-cmp) plugin to be installed (and properly configured).
+
+```lua
+-- Registers copilot-chat source and enables it for copilot-chat filetype (so copilot chat window)
+require("CopilotChat.integrations.cmp").setup()
+
+-- You might also want to disable default <tab> complete mapping for copilot chat when doing this
+require('CopilotChat').setup({
+  mappings = {
+    complete = {
+      insert = '',
+    },
+  },
+  -- rest of your config
+})
+```
+
+![image](https://github.com/CopilotC-Nvim/CopilotChat.nvim/assets/5115805/063fc99f-a4b2-4187-a065-0fdd287ebee2)
+
+</details>
+
 ## Roadmap (Wishlist)
 
 - Use indexed vector database with current workspace for better context selection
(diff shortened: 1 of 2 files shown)


taketwo:
Any hints regarding where `require("CopilotChat.integrations.cmp").setup()` should be called? As part of `cmp` setup? Or as part of `CopilotChat` setup?

deathbeam:
> Any hints regarding where `require("CopilotChat.integrations.cmp").setup()` should be called? As part of `cmp` setup? Or as part of `CopilotChat` setup?

As part of copilot chat setup ideally but it doesnt rly matter when you call it as long as both plugins will get loaded at some point as all this does is registers the source and filetype to nvim-cmp so it do not requires calling setup on either copilot chat or nvim cmp beforehand.

taketwo:
Got it. I added it as a part of `CopilotChat` setup with `lazy` and it works, like so:

```lua
    config = function(_, opts)
      require('CopilotChat').setup(opts)
      require('CopilotChat.integrations.cmp').setup()
    end,
```

GitMurf:
@deathbeam with the new nvim-cmp integration when you select/accept an item from the cmp list it lists / unhides the buffer.

image

GitMurf:
You should be able to add something like the following... but keep in mind it still will flash as listed for a split second and then unlist again. So maybe you can find a better way 🤷‍♂️ 

```lua
---Executed after the item was selected.
---@param completion_item lsp.CompletionItem
---@param callback fun(completion_item: lsp.CompletionItem|nil)
function Source:execute(completion_item, callback)
  callback(completion_item)
  -- reset the buffer listed to false
  vim.api.nvim_set_option_value('buflisted', false, { buf = 0 })
end
```

deathbeam:
> You should be able to add something like the following... but keep in mind it still will flash as listed for a split second and then unlist again. So maybe you can find a better way 🤷‍♂️
> 
> ```lua
> ---Executed after the item was selected.
> ---@param completion_item lsp.CompletionItem
> ---@param callback fun(completion_item: lsp.CompletionItem|nil)
> function Source:execute(completion_item, callback)
>   callback(completion_item)
>   -- reset the buffer listed to false
>   vim.api.nvim_set_option_value('buflisted', false, { buf = 0 })
> end
> ```

Hmm thats pretty annoying. But split second is probably fine, better than it staying there. Not sure if its worth trying to find the root cause as it sounds like nvim-cmp bug, so your workaround should be good enough for now. Do you want to PR it?

GitMurf:
I will submit a PR.

Also I have some notes that said I researched it a bit and it is an upstream thing from nvim-cmp that is outside of our control... FYI.

GitMurf:
@deathbeam PR submitted: #346

PR #348, opened 2024-05-31
fix: Improve regexp for command completion in chat

festeh:
Hi! 

This MR fixes this bug:

When user types `/Fi` in chat and then presses `<TAB>`
Expected: only `Fix` and `FixDiagnostic` are shown
Result: all actions are shown

Reference: https://neovim.io/doc/user/pattern.html#search-pattern

### DIFF
--- a/lua/CopilotChat/init.lua
+++ b/lua/CopilotChat/init.lua
@@ -135,7 +135,7 @@ local function complete()
     return
   end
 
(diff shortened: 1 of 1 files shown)


deathbeam (approved):

deathbeam:
@all-contributors add @festeh for code

Issue #354, opened 2024-06-06
Question: is `set cursorline` mandatory?


kjuq:
In the status quo, `cursorline` is set when the `copilot-chat` buffer opens. However I'm curious whether it is a necessary behavior to make this plugin works properly. In my case, I prefer no `cursorline` because it is a bit annoying to focus coding.

If it weren't a mandatory option. I made a PR to add an option to disable that. So I will be so happy for you to review that.

Edit:
Sent PR (https://github.com/CopilotC-Nvim/CopilotChat.nvim/pull/355)

deathbeam:
https://github.com/CopilotC-Nvim/CopilotChat.nvim?tab=readme-ov-file#customizing-buffers see

kjuq:
I misunderstood the problem. I opened a new issue (#356).

PR #355, opened 2024-06-06
feat: option to show cursorline or not (#354)

kjuq:
I added an option described at title.
Relates to https://github.com/CopilotC-Nvim/CopilotChat.nvim/issues/354


deathbeam:
See https://github.com/CopilotC-Nvim/CopilotChat.nvim?tab=readme-ov-file#customizing-buffers

PR #368, opened 2024-07-16
Allow runtime selection of models with `:CopilotChatModels`

### LINKED ISSUE #366: feat: gpt-4o support
Hi,
copilot chat now supports gpt-4o. Could you add its support? - I think it would need to use https://openaipublic.blob.core.windows.net/encodings/o200k_base.tiktoken

Thanks.

gptlang:
CC @deathbeam & CC @jellydn 

I haven't touched Lua in a while and don't really know how the codebase works anymore. Could you guys check if I did anything wrong

Tentatively closes #366

### DIFF
--- a/lua/CopilotChat/copilot.lua
+++ b/lua/CopilotChat/copilot.lua
@@ -504,39 +504,43 @@ end
 --- Fetch & allow model selection
 ---@param callback fun(table):nil
 function Copilot:select_model(callback)
-  if self.models_cache == nil then
-    local url = 'https://api.githubcopilot.com/models'
-    self:with_auth(function()
-      local headers = generate_headers(self.token.token, self.sessionid, self.machineid)
-      curl.get(url, {
-        headers = headers,
-        proxy = self.proxy,
-        insecure = self.allow_insecure,
-        on_error = function(err)
-          err = 'Failed to get response: ' .. vim.inspect(err)
-          log.error(err)
-        end,
-        callback = function(response)
-          if response.status ~= 200 then
-            local msg = 'Failed to fetch models: ' .. tostring(response.status)
-            log.error(msg)
-            return
-          end
-
-          local models = vim.json.decode(response.body)['data']
-          local selections = {}
-          for _, model in ipairs(models) do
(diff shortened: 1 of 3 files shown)


pidgeon777:
So, to clarify, it will be possible to dynamically choose which model to use for any given request?

gptlang:
Well, you'll have to specify it. It's not automagic like vscode. It's currently a bit slow due to fetching the models on each request. We might want to cache it per session.

gptlang:
But the point is to make discovery of available models easier, for example to check if gpt-4o is available to you

jellydn:
Hi, @gptlang What if we save the selected model and use that config for next time? It's similar to what `Copilot Auth` does. So we don't need to call models every time. What do you think?

gptlang:
I don't want to touch `init.lua` and would rather have that be followed by default. If you want to use a model by default, set it there

jellydn (approved):
LGTM!

lua/CopilotChat/init.lua
```diff
@@ -354,11 +354,25 @@ function M.toggle(config, source)
...
 end
 
--- @returns string
+--- @returns string
 function M.response()
   return state.response
 end
 
+function M.select_model()
+  state.copilot:select_model(function(model)
```
deathbeam, line 363: this should only change M.config, and the config itself shouldnt ever be nil, it shouldnt change stateconfig.model or default_config.model

lua/CopilotChat/copilot.lua
```diff
@@ -499,6 +501,55 @@ function Copilot:ask(prompt, opts)
   end, on_error)
 end
 
+--- Fetch & allow model selection
+---@param callback fun(string):nil
+function Copilot:select_model(callback)
+  if self.models_cache ~= nil then
```
deathbeam, line 507: you can reduce the code duplication here by making inner local function that does the vim.ui.select. but also i think the select shouldnt be here at all, it should be in the select_model method in init.lua, this should simply call the callback with the model_cache

deathbeam (changes requested):

PR #377, opened 2024-07-26
Use gpt-4o by default

gptlang:
Potentially debatable: automatically rewrite `gpt-4o` to `gpt-4o-2024-05-13` because the server doesn't do that for us despite allowing the short name for `gpt-4` and `gpt-4-turbo`.


jellydn:
Thanks @gptlang LGTM!

PR #384, opened 2024-08-03
Improve completion of Copilot keywords

guill:
Previously, completion in nvim-cmp would only happen when typing the `@` or `/` character. There
was no way to type a partial command (e.g. `/COPILOT_G`) and then autocomplete.

This PR adds a keyword pattern to allow this sort of completion in a case-insensitive way.

If the previous behavior was an intentional design decision, feel free to close this PR. (I changed
it for my own usage, but figured other people might like it as well.)

### DIFF
--- a/lua/CopilotChat/integrations/cmp.lua
+++ b/lua/CopilotChat/integrations/cmp.lua
@@ -7,25 +7,29 @@ function Source:get_trigger_characters()
   return { '@', '/' }
 end
 
+function Source:get_keyword_pattern()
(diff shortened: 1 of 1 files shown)


jellydn:
Thanks @guill. I don't use this feature much so let's see @deathbeam's view on this approach.

deathbeam:
I dont use it either (as I dont use nvim-cmp) but looks good to me

deathbeam (approved):

jellydn:
@all-contributors add @guill for code.

Issue #364, opened 2024-07-04
Diff yank with 'gy' (the default mapping) doesn't work.


marwan-tanager:
Hi, I couldn't get the `gy` mapping to yank the diff in the chat window to work. Also, it doesn't work if I assign any other key combination to `mappings.yank_diff`.

sjonpaulbrown-cc:
I was experiencing the same issue. I had my clipboard set to `vim.o.clipboard = 'unnamedplus'`. When commenting this out, I was able to successfully use the `gy` mapping. 

The `gy` mapping stores the diff to the `"` register. You can validate that the diff is properly getting copied by running the `:registers "` command. After copying with `gy`, you need to validate the contents of the `"` register. If it is empty, you may be overwriting the register in some way.

deathbeam:
Looks like this is resolved by register being configurable and original issue was mostly misunderstanding registers anyway I would say

PR #389, opened 2024-08-08
Add support for overriding the default yank_diff register

sjonpaulbrown-cc:
## Description
The following PR adds a config option to override the default `yank_diff` register. Currently, the default is set to `"`, and it can conflict with configuration settings that set the system clipboard to the unnamed register. This override enables me to set the `yank_diff_register` to the `+` register and maintain my current system clipboard setting of `vim.o.clipboard = 'unnamedplus'`

The default would remain the same, but it would allow this to be changed as needed.

This could address issue #364

### DIFF
--- a/README.md
+++ b/README.md
@@ -199,6 +199,8 @@ Also see [here](/lua/CopilotChat/config.lua):
   proxy = nil, -- [protocol://]host[:port] Use this proxy
   allow_insecure = false, -- Allow insecure server connections
 
+  yank_diff_register = '"', -- Allow overriding the register for yanking diffs
+
   system_prompt = prompts.COPILOT_INSTRUCTIONS, -- System prompt to use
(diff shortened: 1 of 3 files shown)


gptlang (approved):
lgtm

jellydn:
Thanks @sjonpaulbrown-cc 

@all-contributors add @sjonpaulbrown-cc for code.

deathbeam:
This should've been just config on the yank_diff mapping directly no? Or at least it would make more sense there

sjonpaulbrown-cc:
@deathbeam , my initial thought process was to have this be a "global" config setting that could be used for all yank processes rather than for a single mapping. That said, the config setting would have been better named to `yank_register` rather than `yank_diff_register` if it were to be used in that way.

Either way, feel free to move it to the mapping config if it makes more sense to keep it there.

Issue #381, opened 2024-07-29
Seems that the Custom Prompts using a custom system_prompt is not actually changing the system prompt


GitMurf:
The instructions here: https://github.com/CopilotC-Nvim/CopilotChat.nvim?tab=readme-ov-file#custom-system-prompts

show that you can add a custom system_prompt like this:

```lua
{
  system_prompt = 'Your name is Github Copilot and you are a AI assistant for developers.',
  prompts = {
    MyCustomPromptWithCustomSystemPrompt = {
      system_prompt = 'Your name is Johny Microsoft and you are not an AI assistant for developers.',
      prompt = 'Explain how it works.',
    },
  },
}
```

But when I try this with something obvious like the following:

```lua
{
    prompts = {
      MyCustomPiratePrompt = {
        system_prompt = [[You are fascinated by pirates, so please respond in pirate speak.]],
        prompt = 'what is a sorting algo?',
      },
    },
}
```

I get a regular response which appears to be using the main system prompt as the response is NOT in pirate speak. Here is the result:

![image](https://github.com/user-attachments/assets/75fc66b5-de33-468d-bb28-2949c89374c2)

If I override the default system prompt using the following in config:

```lua
{
    ...
    system_prompt = [[You are fascinated by pirates, so please respond in pirate speak.]],
    prompts = { ... },
    ...
}
```

Then I indeed get a response in "pirate speak" confirming the system_prompt works as defined. I would expect the same result for my custom prompt above where using that individual custom prompt would respond in pirate speak just for that specific prompt (and regular chat prompt would use the default system prompt still).

![image](https://github.com/user-attachments/assets/7188c6d7-77d6-4bb2-a691-4f9d68a66f05)

GitMurf:
Also to confirm I turned on the debug flag so that I could see the prompt being used and it indeed is showing the default system prompt and not the custom one.

gptlang:
GitHub not sending me email notifications for some reason. Will look into this

gptlang:
Looks like a documentation issue. 

After defining custom prompts, you can use them by doing `:CopilotChat<whatever_you_named_the_prompt>`

GitMurf:
@gptlang what do you mean? I did not have a problem executing my custom prompt (using "/MyCustomPrompt" in the message) and it correctly runs the prompt EXCEPT not if I try to override the "system" prompt with it. It just appears that the system prompt is not being customized. Does that make sense?

[...]

PR #394, opened 2024-08-18
fix(readme): unnecessary [[brackets]] around pirate system prompt example.

GitMurf:


GitMurf:
@gptlang I believe you  copied this example from my gh issue #381 but I had used [[for encapsulating my string]] so do not need both quotes and the [[string brackets]]. Don’t want it to confuse users if they try and figure out why there are brackets in there inside string quotes.

gptlang (approved):

PR #400, opened 2024-08-29
Cleanup auto insert behaviour and add support for insert at end

### LINKED ISSUE #379: Why isn't a prompt buffer used?
One thing I find annoying is that after I scroll up to view the history, entering insert mode then allows me to edit the current line (e.g. history) instead of taking me back to the prompting line/area. It'd be nice if entering insert mode took me back to the current prompt automatically (and IIRC, this is precisely how prompt buffers work in neovim).

Here's my temporary workaround, although there may be a better way to implement this internally (using a prompt buffer (see `:help prompt-buffer`), or, extmarks, etc.):

```lua
vim.api.nvim_create_autocmd('FileType', {
  group = augroup,
  -- pattern = 'copilot-*',
  pattern = 'copilot-chat',
  callback = function(info)
    vim.api.nvim_create_autocmd('InsertEnter', {
      group = augroup,
      buffer = info.buf,
      desc = 'Move to last line',
      -- nested = true,
      callback = function()
        vim.cmd 'normal! 0'

        if vim.fn.search([[\m^##\s\+\S]], 'cnW') > 0 then
          vim.cmd 'normal! G$'
          vim.v.char = 'x'
        end
      end,
    })
  end,
})
```

deathbeam:
- Cleanup auto insert behaviour and move it to chat.lua
- Add support for moving cursor to end of buffer before entering insert mode in the chat buffer
- Cleanup configuration (move yank register config near the mapping)

Closes #379

### DIFF
--- a/README.md
+++ b/README.md
@@ -201,8 +201,6 @@ Also see [here](/lua/CopilotChat/config.lua):
   proxy = nil, -- [protocol://]host[:port] Use this proxy
   allow_insecure = false, -- Allow insecure server connections
 
-  yank_diff_register = '"', -- Allow overriding the register for yanking diffs
-
   system_prompt = prompts.COPILOT_INSTRUCTIONS, -- System prompt to use
   model = 'gpt-4o', -- GPT model to use, 'gpt-3.5-turbo', 'gpt-4', or 'gpt-4o'
   temperature = 0.1, -- GPT temperature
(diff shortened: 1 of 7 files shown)


README.md
```diff
@@ -201,8 +201,6 @@ Also see [here](/lua/CopilotChat/config.lua):
   proxy = nil, -- [protocol://]host[:port] Use this proxy
   allow_insecure = false, -- Allow insecure server connections
 
-  yank_diff_register = '"', -- Allow overriding the register for yanking diffs
```
jellydn, line 204: Thanks, @deathbeam Could we add a note and changelog or at least the warning if someone is still using this config?

jellydn (commented):
Nice work, thanks @deathbeam.

README.md
deathbeam: Alright added warning

jellydn (approved):
LGTM!

PR #401, opened 2024-08-31
Move help to separate window

### LINKED ISSUE #361: Show help in separate popup when `?` pressed in normal mode
Having mappings displayed is very useful but current implementation takes a lot of valuable screen estate. 
Many other plugins like [outline](https://github.com/hedyhli/outline.nvim) provide a `?` mapping in normal mode that shows all mappings in separate popup.

We can still show just a single line (configurable) like: "Press '?' for help"

deathbeam:
Closes #361

image

image


jellydn (approved):
LGTM!

PR #413, opened 2024-09-17
Fix indexing nil `config`

fjchen7:
Fix issue that running `:CopilotChatLoad <name>?` without opening any panel before will result in error of indexing nil `config`.


deathbeam:
@all-contributors add @fjchen7 for code

Issue #433, opened 2024-10-29
Anthropic’s Claude 3.5 Sonnet, Google’s Gemini 1.5 Pro, and OpenAI’s o1-preview and o1-mini are coming to GitHub Copilot


pidgeon777:
I would like to announce the following big news:

https://github.blog/news-insights/product-news/bringing-developer-choice-to-copilot/

Currently, the `o1-mini` and `o1-preview` models are available within the existing GitHub Copilot subscriptions. These models will hopefully be supported in `CopilotChat.nvim`. For more details, refer to the related issue [here](https://github.com/CopilotC-Nvim/CopilotChat.nvim/issues/419). We are hopeful that support for the remaining models will also be integrated in the future.

gptlang:
Error for Claude: 
```json
{"error":{"message":"model access is not permitted per policy settings","param":"","code":"","type":""}}
```

gptlang:
Claude support might actually get me back into using Copilot Chat. Anthropic has horrible rate limits but the best models.

pidgeon777:
I also have this:

```
Failed to get response: {"error":{"message":"model access is not permitted per policy settings","param":"","code":"","type":""}}
```

List of available models:

![image](https://github.com/user-attachments/assets/3668a426-a09b-48ac-93b1-22fd08d6956b)

maybe we're missing some input request arguments, or usage is strictly allowed for VSCode for now?

gptlang:
Hmm that's weird. Did you update? It just worked for me

gptlang:
Ok I might have found a reason. Since requests are async, the completion request might arrive before the policy request.

If you try again, it should work since only 1 policy request is ever needed

jellydn:
Great work @gptlang I have just tested with the latest canary and it works fine for me.

snoord:
FYI if you have access to Copilot through a business subscription, this error message is returned on every response when setting the model to `claude-3.5-sonnet`:

```json
{"error":{"message":"cannot enable policy inline for business users","param":"","code":"","type":""}}
```

snoord:
@deathbeam Just tried the new changes, got the following error:

```
Error executing luv callback:
...e/nvim/lazy/CopilotChat.nvim/lua/CopilotChat/copilot.lua:401: bad argument #1 to 'find' (string expected, got nil)
stack traceback:
        [C]: in function 'find'
[...]

PR #435, opened 2024-10-29
Claude Support

gptlang:
Closes:
- #433


gptlang:
Using claude requires explicitly enabling the feature in vscode. We just send the same request and boom

PR #455, opened 2024-11-04
Improve token counting and use token count from github response

deathbeam:


deathbeam:
Added automatic history trimming as well on nearing the token limits

deathbeam:
image

also added token max count display now that we are fetching it from github

PR #460, opened 2024-11-06
fix: "Unterminated quoted string" in fzflua; extra quotes in actions prompts

liskin:
#### [fix(fzflua): Syntax error: Unterminated quoted string](../commit/519a654b1eb185505092c488b6271d68d02e7065)

The help_actions prompts end with a double quote, and passing that to
shell unescaped explodes with

    sh: 1: Syntax error: Unterminated quoted string

We could escape it using fzf-lua.libuv.shellescape, but it's easier to
just let fzf-lua do the right thing itself:
https://github.com/ibhagwan/fzf-lua/blob/052491ee887b04e231787515a81aa3f358a68147/lua/fzf-lua/core.lua#L645-L646

#### [fix(actions): Drop unmatched double quotes from prompts](../commit/4d087d39518a829bf8761f200f3322fbfd425904)

There's no point having a quote at the end of the prompt if it's never
getting closed (and it isn't indeed). Also, the prompt used for
CopilotChatFixDiagnostic doesn't have any quotes either, so to be
consistent let's just remove them from actions prompts as well.

### DIFF
--- a/lua/CopilotChat/actions.lua
+++ b/lua/CopilotChat/actions.lua
@@ -24,11 +24,11 @@ function M.help_actions(config)
     prompt = 'Copilot Chat Help Actions',
     actions = {
       ['Fix diagnostic'] = vim.tbl_extend('keep', {
-        prompt = 'Please assist with fixing the following diagnostic issue in file: "',
+        prompt = 'Please assist with fixing the following diagnostic issue in file:',
         selection = select.diagnostics,
       }, config or {}),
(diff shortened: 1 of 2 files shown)


liskin:
(I can split this into 3 PRs if that's preferable)

deathbeam:
Nah its fine. I had these small grievances with the fzf lua integration for a while as well but never actually got time for fixing it so this is great, thanks!

PR #467, opened 2024-11-10
Automatically sanitize input sent to copilot API

### LINKED ISSUE #464: claude-3.5-sonnet returns 500 on some special unicode characters
When i set `model = "claude-3.5-sonnet"` and select my whole buffer or set `context = "buffers"` and ask something to Copilot while having claude as selected model, i have this error that appears :

```
Failed to parse response: "Expected value but found invalid token at character 1"
Internal Server Error
```

Debug logs :
```
[CopilotChat.nvim] [DEBUG 00:41:04]
~/.local/share/nvim/lazy/CopilotChat.nvim/lua/CopilotChat/context.lua:239: Got 1 embeddings 

[CopilotChat.nvim] [DEBUG 00:41:05]
~/.local/share/nvim/lazy/CopilotChat.nvim/lua/CopilotChat/copilot.lua:499: Temperature: 0.1 

[CopilotChat.nvim] [DEBUG 00:41:08]
~/.local/share/nvim/lazy/CopilotChat.nvim/lua/CopilotChat/copilot.lua:513: Tokenizer: o200k_base

[CopilotChat.nvim] [INFO  00:41:08]
~/.local/share/nvim/lazy/CopilotChat.nvim/lua/CopilotChat/copilot.lua:468: Claude enabled 

[CopilotChat.nvim]
Internal Server Error
```

Here is my current CopilotChat config :
```
{
    event = "BufReadPre",
    branch = "canary",
    dependencies = {
        { "github/copilot.vim" },
        { "nvim-lua/plenary.nvim" },
    },
    build = "make tiktoken",
    opts = {
        debug = true,
        window = {
            layout = "float",
            border = "rounded",
            title = "Copilot Chat",
        },
        model = "claude-3.5-sonnet",
        question_header = "  User ",
        answer_header = "  Copilot ",
        error_header = "  Error ",
        auto_insert_mode = true,
        insert_at_end = true,
        context = "buffers",
        highlight_selection = false,
    },
    config = function(_, opts)
        local chat = require("CopilotChat")
        chat.setup(opts)

        local wk = require("which-key")
        wk.add({
            {
                group = "Copilot",
                "<leader>g",
                { "<leader>cc", chat.toggle, mode = { "n", "v" }, desc = "Toggle Copilot Chat", icon = { icon = "󰠠", color = "white" } },
                { "<leader>ca", ask_copilot, mode = { "n", "v" }, desc = "Ask Copilot", icon = { icon = "󰠠", color = "white" } },
                icon = { icon = "󰠠", color = "white" },
            },
        })
    end,
},
```

deathbeam:
Removes all non-standard characters to avoid issues with copilot API (especially claude).

Closes #464

### DIFF
--- a/lua/CopilotChat/copilot.lua
+++ b/lua/CopilotChat/copilot.lua
@@ -74,6 +74,31 @@ local function machine_id()
   return hex
 end
 
+local function sanitize_text(model, text)
+  if not text then
+    return ''
+  end
+
+  if not vim.startswith(model, 'claude') then
+    return text
+  end
+
+  -- Remove various emoji, symbols, and special characters:
+  return text:gsub(
+    '[\u{1F000}-\u{1F9FF}' -- Extended emoji and symbols
+      .. '\u{2600}-\u{26FF}' -- Miscellaneous symbols
+      .. '\u{2700}-\u{27BF}' -- Dingbats
+      .. '\u{FE00}-\u{FE0F}' -- Variation Selectors
+      .. '\u{1F1E6}-\u{1F1FF}' -- Regional indicator symbols
+      .. '\u{1F300}-\u{1F5FF}' -- Miscellaneous Symbols and Pictographs
+      .. '\u{1F600}-\u{1F64F}' -- Emoticons
+      .. '\u{1F680}-\u{1F6FF}' -- Transport and Map Symbols
+      .. '\u{1F900}-\u{1F9FF}' -- Supplemental Symbols and Pictographs
+      .. '\u{E000}-\u{F8FF}]', -- Private Use Area
+    ''
+  )
+end
+
 local function find_config_path()
   local config = vim.fn.expand('$XDG_CONFIG_HOME')
   if config and vim.fn.isdirectory(config) > 0 then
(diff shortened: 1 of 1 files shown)


GitMurf:
One thought @deathbeam is that maybe we want to make this a config option to opt into? or at least have a way to disable if there is a reason a user wants to send icons / emojis. For example, I actually do sometimes ask about emojis / icons for different configs for neovim plugins (other other similar stuff). "Can you provide me some alternative icons or emojis for 📣 ?" 

Another though is that I have NEVER experienced this with any other model other than Claude. So my proposal would be that this sanitization is only done for claude (at least to start). You then could probably get away with not providing the suggested config item above if it was a documented requirement for Claude due to its issues with certain characters. Although I would still recommend a config item so that users aren't caught off guard why all the sudden certain characters are being removed "randomly".

Lastly I have no clue, but would this affect other non english languages in more significant ways (again why I would suggest a config item)?

deathbeam:
I dont like config option for something that just breaks the API. Like if the API do not works with some emojis then enabling asking about them doesnt rly achieve anything either right. But I guess we can just limit it to claude as other models dont have this issue. Also good point about other languages as I guess this would also filter out chinese and other characters.

GitMurf:
> I dont like config option for something that just breaks the API. Like if the API do not works with some emojis then enabling asking about them doesnt rly achieve anything either right. But I guess we can just limit it to claude as other models dont have this issue. Also good point about other languages as I guess this would also filter out chinese and other characters. 

Yeah then if someone has an issue with characters they need (or language) then the solution can be to just use a non Claude model.

PR #490, opened 2024-11-15
Add support for copilot extension agents

### LINKED ISSUE #466: Copilot Extentions agent Support
GitHub recently introduced Copilot Extensions:
https://docs.github.com/en/copilot/building-copilot-extensions/about-building-copilot-extensions

For instance, if a user enables the [Perplexity AI extension](https://github.com/marketplace/perplexityai), they can send requests to the Perplexity API directly from Copilot.

It would be extremely useful if CopilotChat.nvim could support calling these agents as well. In VS Code, this can be done by using the `@` prefix, such as `@perplexityai`.
Screenshot 2024-11-05 10:23:23

In my testing with Neovim, I found that Perplexity can be called with the Copilot token by implementing this approach:
https://github.com/CopilotC-Nvim/CopilotChat.nvim/pull/465
Screenshot 2024-11-10 14:56:48

In the case of Perplexity, responses include content.copilot_references, so displaying referenced URLs and titles could enhance the user experience. I believe this addition would integrate seamlessly with the current implementation and add significant convenience!

In addition, sending a GET request to `https://api.githubcopilot.com/agents` retrieves a list of agents available to the user.

deathbeam:
https://docs.github.com/en/copilot/building-copilot-extensions/about-building-copilot-extensions

- Change @buffer and @buffers to #buffer and #buffers
- Add support for @agent agent selection
- Add support for config.agent for specifying default agent
- Add :CopilotChatAgents for listing agents (and showing selected agent)
- Remove :CopilotChatModel, instead show which model is selected in :CopilotChatModels
- Remove early errors from curl so we can actually get response body for the error
- Add info to README about models, agents and contexts

Closes #466

image

### DIFF
--- a/README.md
+++ b/README.md
@@ -110,7 +110,7 @@ Verify "[Copilot chat in the IDE](https://github.com/settings/copilot)" is enabl
 - `:CopilotChatLoad <name>?` - Load chat history from file
 - `:CopilotChatDebugInfo` - Show debug information
 - `:CopilotChatModels` - View and select available models. This is reset when a new instance is made. Please set your model in `init.lua` for persistence.
-- `:CopilotChatModel` - View the currently selected model.
+- `:CopilotChatAgents` - View and select available agents. This is reset when a new instance is made. Please set your agent in `init.lua` for persistence.
 
 #### Commands coming from default prompts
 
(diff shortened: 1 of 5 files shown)


deathbeam:
Also improved unauthorized message:

image

deathbeam:
Example of using @models agent:

image

PR #494, opened 2024-11-16
feat: Add support for github marketplace models

deathbeam:
Adds integration with Github Marketplace models available through Azure ML API.
The models are fetched from the Azure catalog and exposed through the same
interface as regular Copilot models. This provides users with access to
additional experimental models while maintaining consistent UX.

The docs are updated to point to official Copilot model documentation and
marketplace models page rather than listing models inline.


00sapo:
Hello, it's not clear to me how to use models from the marketplace. Could you explain it?

deathbeam:
Use `CopilotChatModels` command and they should be there with `github_models` in the name

00sapo:
image
This is what I see now

deathbeam:
that looks like outdated display if you used `CopilotChatModels`, update your plugin. but also if you dont see github models, then you cant access them (happens if you have copilot through organization for example).

PR #495, opened 2024-11-16
Add support for #files (workspace file map) context

deathbeam:
Support workspace file map context. This context will simply include all related filenames to the prompt in chat context. Useful for searching where something might be (but limited to only searching on filenames for now).

image

image


deathbeam:
image

debug info also contains file map now

PR #497, opened 2024-11-16
feat(chat): improve auto follow cursor behavior

taketwo:
Auto follow cursor is generally a convenient and desirable feature. However, it gets annoying when the model is streaming a very long response spanning more vertical space than the chat window has. In such cases the user might want to read the response from the beginning at a human pace, yet auto follow keeps pushing forward at robot speed.

This PR improves auto follow cursor behavior by conditioning it on the current cursor position. If it is at the last line, auto-following happens. Otherwise, the cursor stays where it is. Effectively, this means that the user can prevent auto-following from happening by simply moving the cursor up.

### DIFF
--- a/lua/CopilotChat/chat.lua
+++ b/lua/CopilotChat/chat.lua
@@ -2,7 +2,6 @@
 ---@field bufnr number
 ---@field winnr number
 ---@field separator string
----@field auto_follow_cursor boolean
 ---@field spinner CopilotChat.Spinner
 ---@field valid fun(self: CopilotChat.Chat)
 ---@field visible fun(self: CopilotChat.Chat)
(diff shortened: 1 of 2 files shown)


lua/CopilotChat/chat.lua
```diff
@@ -129,6 +131,16 @@ function Chat:append(str)
     self.spinner:start()
   end
 
+  -- Decide if we should follow cursor after appending text.
+  -- Note that winnr may be nil if the chat window is not open yet.
+  local should_follow_cursor = self.auto_follow_cursor
+  if self.auto_follow_cursor and self.winnr then
```
deathbeam, line 137: this check should probably be behind self:visible() instead of just checking if winnr is set.

deathbeam (changes requested):

lua/CopilotChat/chat.lua
```diff
@@ -2,6 +2,7 @@
 ---@field bufnr number
 ---@field winnr number
 ---@field separator string
+---@field auto_follow_cursor boolean
```
deathbeam, line 5: dont need to add this here I think, lsp should know by default when working in this file, and when working outside we dont want this field "exposed" anyway

deathbeam (changes requested):

taketwo (commented):
Thanks for the review, makes sense, I've pushed a fixup commit.

deathbeam (approved):
Looks good, thanks!

deathbeam:
@all-contributors add @taketwo for code

PR #507, opened 2024-11-17
feat: add automatic chat autocompletion

deathbeam:
Adds new config option `chat_autocomplete` (enabled by default) that will trigger completion automatically when typing trigger characters (@, /, #, $, :). This replaces the nvim-cmp integration which is now deprecated in favor of this new solution.

The automatic completion is debounced to avoid too many triggers and uses the same completion logic as before, just moved to separate function.

### DIFF
--- a/README.md
+++ b/README.md
@@ -276,6 +276,7 @@ Also see [here](/lua/CopilotChat/config.lua):
   error_header = '## Error ', -- Header to use for errors
   separator = '───', -- Separator to use in chat
 
+  chat_autocomplete = true, -- Enable chat autocompletion (when disabled, requires manual `mappings.complete` trigger)
   show_folds = true, -- Shows folds for sections in chat
   show_help = true, -- Shows help message as virtual lines when waiting for user input
   auto_follow_cursor = true, -- Auto-follow cursor in chat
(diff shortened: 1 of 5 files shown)


dfsnow:
Is it possible to make this change optional i.e. keep the `cmp` integration but have it off by default? The new built-in completion behavior does fully replicate `cmp` and it trips me up constantly.

As an example, hitting escape while in `cmp` will remove the completed text and return to insert mode. Hitting escape in the built-in completion leaves the completed text and returns to normal mode.

deathbeam:
> Is it possible to make this change optional i.e. keep the `cmp` integration but have it off by default? The new built-in completion behavior does fully replicate `cmp` and it trips me up constantly.
> 
> As an example, hitting escape while in `cmp` will remove the completed text and return to insert mode. Hitting escape in the built-in completion leaves the completed text and returns to normal mode.
> 
> The new built-in also prevents completion from other sources within `copilot-*` buffers.

I think what we have now makes more sense than trying to support nvim-cmp and every other completion plugin that ppl would ask for. This was mostly in response to bunch of new popular completion plugins coming out and I think simply providing api for easy integration like the plugin does now and providing sensible default implementation is better than trying to integrate everything. And the behaviour of the default vim completion menu is configurable as well so im sure theres some stuff you can do with it. 

I guess we could also maybe set omnifunc for the buffer and then you can integrate it with other completion providers even more easily, i might look at that as well, but other than that I think the current implementation is fine.

PR #563, opened 2024-11-20
Try to force http1.1 and tls1.2 for windows

deathbeam:


deathbeam:
This wasnt the solution after all so closing

PR #568, opened 2024-11-20
feat: add register context and improve selection handling

deathbeam:
Adds new register context that allows accessing vim register contents in chat
and improves selection handling by:

- Making selection fields required instead of optional
- Improving unnamed register selection to include buffer context
- Deprecating clipboard selection in favor of register context
- Adding proper mark setting for better selection tracking
- Supporting array of default contexts in config

The register context supports all standard vim registers and provides a UI
for selecting them when used. This makes accessing clipboard and other
registers more intuitive.

### DIFF
--- a/README.md
+++ b/README.md
@@ -240,7 +240,7 @@ What is 1 + 11
 ### Models
 
 You can list available models with `:CopilotChatModels` command. Model determines the AI model used for the chat.  
-You can set the model in the prompt by using `$` followed by the model name.  
+You can set the model in the prompt by using `$` followed by the model name or default model via config using `model` key.  
 Default models are:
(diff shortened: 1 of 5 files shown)


jellydn (approved):
LGTM!

@deathbeam Could you update this usage on README as well?

deathbeam:
> LGTM!
> 
> @deathbeam Could you update this usage on README as well?

updated API section a bit but not sure what else needs updating, i added it to README to contexts and rest of context stuff is already explained there I think

deathbeam:
Ok swapped unnamed register back to selection as we can get position info for it via `[` and `]`

PR #682, opened 2024-12-09
fix: refactor ask function to handle content properly

atkodev:
Hi there! First, thank you for writing this plugin. I'm using it well. 

But I found that the o1 model returns nothing, with an empty response error message. So I fixed it. Could you please review this PR and consider merging it? I haven't contributed much before, so I apologize in advance if I've made any mistakes. Thank you!


deathbeam:
looks good thanks.

@all-contributors add @atkodev for code

PR #708, opened 2025-01-06
docs: update README with setup function for lazy installation

arjunmahishi:
I think d82c5a30571e9759335cc50a27c0de0573e888f7 removed the `setup()` function from the lazy.nvim installation snippet. `setup()` seems to be mandatory for this plugin as all the user commands are created within it.

### DIFF
--- a/README.md
+++ b/README.md
@@ -51,6 +51,11 @@ return {
       -- See Configuration section for options
     },
     -- See Commands section for default commands if you want to lazy load on them
+    config = function()
+      require("CopilotChat").setup {
+        -- See Configuration section for options
+      }
+    end
   },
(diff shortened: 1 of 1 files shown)


jellydn:
Thanks @arjunmahishi however, no need to do so. If we have `opts`, lazy.nvim will call setup. Here is example from https://lazy.folke.io/spec/examples

```lua
 {
    "nvim-neorg/neorg",
    -- lazy-load on filetype
    ft = "norg",
    -- options for neorg. This will automatically call `require("neorg").setup(opts)`
    opts = {
      load = {
        ["core.defaults"] = {},
      },
    },
  },
```

PR #710, opened 2025-01-08
feat: Add theme options to select Telescope themes

Kohei-Wada:
Added theme options to the M.pick function to allow selection of dropdown, ivy, and cursor themes.

image
image

### DIFF
--- a/lua/CopilotChat/integrations/telescope.lua
+++ b/lua/CopilotChat/integrations/telescope.lua
@@ -13,21 +13,16 @@ local M = {}
 --- Pick an action from a list of actions
 ---@param pick_actions CopilotChat.integrations.actions?: A table with the actions to pick from
 ---@param opts table?: Telescope options
----@param theme string?: Telescope theme
-function M.pick(pick_actions, opts, theme)
+function M.pick(pick_actions, opts)
   if not pick_actions or not pick_actions.actions or vim.tbl_isempty(pick_actions.actions) then
     return
   end
 
   utils.return_to_normal_mode()
-  theme = theme or 'dropdown'
-  local theme_functions = {
-    dropdown = themes.get_dropdown,
-    ivy = themes.get_ivy,
-    cursor = themes.get_cursor,
-  }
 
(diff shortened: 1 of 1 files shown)


jellydn:
Thank you @Kohei-Wada for this nice improvement. I like this idea, however, in terms of API (.pick()), I would like to hear @deathbeam thought on this.

deathbeam:
I think better way is to just check if .theme is present in opts or not and if present dont call themes.get_dropdown, see how the themes are impleented in telescope: https://github.com/nvim-telescope/telescope.nvim/blob/415af52339215926d705cccc08145f3782c4d132/lua/telescope/themes.lua

so you could call .pick(actions, telescope.themes.get_ivy(opts)) etc.

Kohei-Wada:
Thank you for your review! I have updated the implementation to allow calling the function as follows:

```lua
actions.prompt_actions({ selection = select.visual }, {
  theme = require("telescope.themes").get_cursor({ prompt_prefix = "copilot-chat>" }),
})
```
This change ensures that the theme can be passed directly in the opts parameter, supporting both default and custom themes provided by the Telescope API. Let me know if you have any further suggestions!

deathbeam:
I did not exactly meant it like that, I meant using theme just as a marker as the .get_cursor sets the `.theme` variable on the opts. So if the variable is set just use the whole opts otherwise use .get_dropdown, as the opts are extended, so you just pass the other opts to the theme function itself:

https://github.com/nvim-telescope/telescope.nvim/blob/415af52339215926d705cccc08145f3782c4d132/lua/telescope/themes.lua#L64

Kohei-Wada:
@deathbeam 
Thank you for the feedback! I think I understood your suggestion to use opts.theme if it is already set and avoid calling themes.get_dropdown. This approach makes sense, as it respects the custom theme provided by the user.

deathbeam (approved):

deathbeam:
Looks good, thanks.

@all-contributors add @Kohei-Wada for code

PR #715, opened 2025-01-16
fix: after inserting a completion remain in insert mode

SamSaffron:
We were not remaining in insert mode making it super easy to submit by mistake

This adds a newline and stays in insert mode


jellydn:
This behavior is by design. I think that's a safe approach. What do you think? @deathbeam

deathbeam:
Im not really sure what problem is this PR solving. I remain in insert mode if I press tab complete already. And I dont like calling complete on opening chat window either that does nothing. If you want to pre-populate agents and models for completion, then just call list_agents and list_models instead

PR #742, opened 2025-02-04
context: Fix bugs.

johncming:
We use files not booleans.

This pull request includes a small change to the `lua/CopilotChat/context.lua` file. The change modifies the `quickfix` function to use `vim.tbl_keys` instead of `vim.tbl_values` when processing `unique_files`.

* [`lua/CopilotChat/context.lua`](diffhunk://#diff-2d1d203284a8d057142c9904421ec216f7b8dafc696915f24f3552608dfb5e75L562-R562): Changed the `quickfix` function to use `vim.tbl_keys` instead of `vim.tbl_values` when iterating over `unique_files`.

PR #744, opened 2025-02-04
context: Fix bugs.

johncming:
It should be “file” here.


deathbeam:
You already did same thing here no? https://github.com/CopilotC-Nvim/CopilotChat.nvim/pull/742/files why are you opening this second time

PR #749, opened 2025-02-05
feat(context): add support for commit-specific git diffs

deathbeam:
Allow viewing diffs for specific commits by passing commit hash/reference to the git context. This extends the existing git context functionality which previously only supported staged and unstaged changes.

The change includes:
- Updated documentation in README.md to reflect new git context options
- Modified context handler to support commit references
- Updated config description to mention commit number support

### DIFF
--- a/README.md
+++ b/README.md
@@ -267,16 +267,23 @@ If context supports input, you can set the input in the prompt by using `:` foll
 Default contexts are:
 
 - `buffer` - Includes specified buffer in chat context. Supports input (default current).
+  - `buffer:<number>` - Includes buffer with specified number in chat context.
 - `buffers` - Includes all buffers in chat context. Supports input (default listed).
+  - `buffers:listed` - Includes only listed buffers in chat context.
+  - `buffers:all` - Includes all buffers in chat context.
 - `file` - Includes content of provided file in chat context. Supports input.
+  - `file:<path>` - Includes content of specified file in chat context.
 - `files` - Includes all non-hidden files in the current workspace in chat context. Supports input (default list).
   - `files:list` - Only lists file names.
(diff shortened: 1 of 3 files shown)


README.md
```diff
@@ -267,16 +267,23 @@ If context supports input, you can set the input in the prompt by using `:` foll
...
 - `file` - Includes content of provided file in chat context. Supports input.
+  - `file:<path>` - Includes content of specified file in chat context.
 - `files` - Includes all non-hidden files in the current workspace in chat context. Supports input (default list).
   - `files:list` - Only lists file names.
   - `files:full` - Includes file content for each file found. Can be slow on large workspaces, use with care.
-- `git` - Requires `git`. Includes current git diff in chat context. Supports input (default unstaged).
+- `git` - Requires `git`. Includes current git diff in chat context. Supports input (default unstaged, also accepts commit number).
   - `git:unstaged` - Includes unstaged changes in chat context.
   - `git:staged` - Includes staged changes in chat context.
+  - `git:<commit>` - Includes changes from specified commit in chat context.
```
liskin, line 282: This is a bit confusing. I expected this to provide the changes in the specified commit itself as a context, but what actually happens is that it provides the output of `git diff <commit>` as context, which is the diff between the current working tree and the specified commit. In other words, it's changes **since** the specified commit. Is that intentional?

I mean, it may well be, I'd just find it more useful to be able to provide `git show <commit>` as a context and — as an example — ask Copilot to look at it and make documentation changes for that commit. 🙂

deathbeam: It is intentional yea, the context is even named git_diff. But you can also just use all the input you normally could use when using git diff so `commit1 commit2` etc, i think its way more flexible like this

liskin: Fair enough.

`#git:<commit>^!` does exactly what I expected.

PR #802, opened 2025-02-18
fix: type annotation of config.providers

unlimitedsola:
I might be wrong but I believe the type annotation for this one was a mistake based on the types from other fields.

Noticed this because I was getting the spurious warning shown below:

image


deathbeam (approved):

deathbeam:
Opps, thanks :d

deathbeam:
@all-contributors add @unlimitedsola for docs and code

PR #804, opened 2025-02-18
feat: Add github workspace command

deathbeam:
Supports resolving github code index workspace data and searching in it

TODO: Currently this api do not accepts ghu_ github copilot token, and I need to use `gh cli` instead that creates hosts.yml

https://github.blog/engineering/the-technology-behind-githubs-new-code-search/


deathbeam:
Well it already works just the separate token is annoying. I was thinking of instead adding it as example as custom context to discussions for now

PR #837, opened 2025-02-21
feat(telescope): add action callback and fix deprecation

chewcw:
It would be convenient if I can do something else with the selection after I pick the action from the Telescope action picker.

Also fixed a deprecated function (vim.api.nvim_win_set_option).

### DIFF
--- a/lua/CopilotChat/integrations/telescope.lua
+++ b/lua/CopilotChat/integrations/telescope.lua
@@ -13,7 +13,8 @@ local M = {}
 --- Pick an action from a list of actions
 ---@param pick_actions CopilotChat.integrations.actions?: A table with the actions to pick from
 ---@param opts table?: Telescope options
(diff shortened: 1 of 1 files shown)


chewcw:
> I dont think doing this only for telescope makes sense. But also I kinda wanted to rework the pickers completely so they can just be configured on plugin instead of calling them manually, which would also improve stuff like #file: pickers etc instead of relying on user to override their vim.ui.select

Cool, do you think you need help for that?

deathbeam:
So in the end decided to just go with relying on `vim.ui.select` completely and instead improve documentation. Could you adjust the PR so it only solves the deprecation? Also added example on how to set callback that you can use as technically the plugin already supports callbacks from the shared config that you can pass around

deathbeam:
Closing because of no reply

PR #853, opened 2025-02-23
Fix: Replace iter with ipairs for Safe Iteration

sgrsydv:
### **PR Description:**  

I was not able to use this plugin inside neovim because of this issue. The user facing error is "model not found: gpt-4o

Replaced `iter` with `ipairs` for safer iteration over API response data. Added validation checks to ensure `response.body.data` and `response.body.summaries` are valid tables before iterating. This prevents runtime errors when the response is `nil` or not structured as expected. Improves stability and error handling without breaking functionality.

Screenshot from 2025-02-23 15-11-36

### DIFF
--- a/lua/CopilotChat/config/providers.lua
+++ b/lua/CopilotChat/config/providers.lua
@@ -156,23 +156,26 @@ M.copilot = {
       error(err)
     end
 
-    local models = vim
-      .iter(response.body.data)
-      :filter(function(model)
-        return model['capabilities']['type'] == 'chat'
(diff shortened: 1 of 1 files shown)


deathbeam:
What is your nvim version? As we support only neovim 0.10.0+, there are some other places that rely on 0.10.0+ features as well

deathbeam:
Closing because of no response and documentation clearly stating we dont support < neovim 0.10.0 which has vim.iter

PR #855, opened 2025-02-24
refactor!: centralize sticky prompt and selection handling

### LINKED ISSUE #716: Making implicit context explicit visually in copilot
I wonder how we would feel about a feature that calls out "current context" more explicitly if set?

So for example if context is a selection  it could say: 

#selection:1:2:33

Then it would be very clear what is going on all the time and you could just edit out implicit contexts if you want to shift gears? 

we could even auto fold in the context, so you could expand to see it.

I know gc finds it just kind of wish I could see it upfront

deathbeam:
This refactoring improves the CopilotChat plugin's state management by:
- Adding `insert_sticky` function to centralize sticky prompt management
- Creating `set_selection` to provide a cleaner API for selection handling
- Consolidating highlight management in `update_highlights`
- Standardizing function signatures and return values
- Improving `Chat:set_prompt` functionality
- Removing redundant code for better maintainability

BREAKING CHANGES:
- `M.resolve_prompt` now returns config first and prompt second
- `M.update_selection` removed in favor of new `M.set_selection` function
- `M.resolve_embeddings` renamed to `M.resolve_context`
- Selection setting now only configurable in global config, not per-operation
- `get_selection` no longer accepts a config parameter
- Selection handling moved from dynamic config to static global config
- Changed sticky format and processing approach
- Prompts in config now use explicit `sticky` property instead of prefixing

Closes #716

### DIFF
--- a/README.md
+++ b/README.md
@@ -453,11 +453,6 @@ Below are all available configuration options with their default values:
   headless = false, -- Do not write to chat buffer and use history(useful for using callback for custom processing)
   callback = nil, -- Callback to use when ask response is received
 
-  -- default selection
-  selection = function(source)
-    return select.visual(source) or select.buffer(source)
-  end,
-
   -- default window options
   window = {
     layout = 'vertical', -- 'vertical', 'horizontal', 'float', 'replace'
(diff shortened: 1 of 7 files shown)


deathbeam:
TODO:

When doing .stop() and starting new prompt right after, for some reason stickies are not inherited

PR #897, opened 2025-03-04
fix(utils): `glob_to_regex` recursion matching

ThisIsMani:
When I use `**/*` pattern in the glob, it doesn't match the files in the subdirectories.

This is because `**/*` is first replaced with `.*` by `:gsub('%*%*/%*', '.*')`.

Then `:gsub('%*', '[^/]*')` kicks in and replaces that `*` with `[^/]*`. So, files in the subdirectories aren't getting matched.

Changed `:gsub('%*', '[^/]*')` to `:gsub('([^%.])%*', '%1[^/]*')`, so that `*` will only get replaced with it's not having a `.` before it.


deathbeam (approved):

deathbeam:
Thanks!

@all-contributors add @ThisIsMani for code

PR #916, opened 2025-03-06
feat: allow per-chat selection configurations

deathbeam:
Make selection behavior configurable per-chat instance rather than only globally. This enables different selection strategies for different chat sessions while maintaining backward compatibility with the global configuration.

The selection field is moved higher in the config file for better visibility, and documentation is updated to reflect this change.

### DIFF
--- a/README.md
+++ b/README.md
@@ -364,7 +364,7 @@ Define your own contexts in the configuration with input handling and resolution
 
 ## Selections
 
-Selections determine the source content for chat interactions. Can be configured only globally.
+Selections determine the source content for chat interactions.
 
 Available selections are located in `local select = require("CopilotChat.select")`:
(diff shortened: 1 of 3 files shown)


liskin:
👍 — this seems super useful for those Commit prompts that should only look at the context (git:staged) and ignore selection. a63031fc706d4e34e118c46339ae2b5681fab21e solves the simple usecase of `:CopilotChatCommit` in a long buffer, but if it's ever reverted, it'd be nice to be able to say `selection = select.empty` or something like that to explicitly disable sending selection as context for a given prompt. Looking through the code, it _seems_ that `selection = function()end` should do the trick but I think that should either be documented or added to `CopilotChat.config.prompts`

PR #1031, opened 2025-03-18
Respect &splitbelow and &splitright

teoljungberg:
This commit changes the split behavior to respect to the value of
&splitbelow and &splitright, which are both disabled by default.

This behavior was introduced in #950.

### DIFF
--- a/lua/CopilotChat/ui/chat.lua
+++ b/lua/CopilotChat/ui/chat.lua
@@ -305,6 +305,8 @@ function Chat:open(config)
     end
     if vim.api.nvim_get_option_value('splitright', {}) then
       cmd = 'botright ' .. cmd
+    else
+      cmd = 'topleft ' .. cmd
     end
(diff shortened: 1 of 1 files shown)


deathbeam:
Shouldnt this then use topleft when splitbelow is not set? and same for splitting vertically with splitright?

teoljungberg:
> Shouldnt this then use topleft when splitbelow is not set? and same for splitting vertically with splitright?

It works with `:split` by default, but I can add topleft to the codepaths to clarify if you want to!

teoljungberg:
Pushed to respect both &splitbelow and &splitright, and to use topleft when either are not set.

What this what you had in mind @deathbeam ?

deathbeam (approved):

deathbeam:
yes this looks good to me

deathbeam:
@all-contributors add @teoljungberg for code

Issue #1052, opened 2025-03-24
Copilot Chat - "Unauthorized" Error


ThomasCode92:
I'm trying to use Copilot Chat but keep getting the following error:

```
Failed to get response: {
  "error": {
    "code": "unauthorized",
    "message": "Your account type is not currently supported",
    "details": "Your account type is not currently supported"
  }
}
```

I have access to Copilot through my work organization, and Copilot Chat in the IDE is enabled.
![Image](https://github.com/user-attachments/assets/64afcb7f-926f-420a-8493-c98e487bc1f5)
When using Copilot Chat in VS Code with the same credentials, it just works fine.

Has anyone encountered this issue before or know how to resolve it?

__Screenshot from within Neovim__
![Image](https://github.com/user-attachments/assets/826266d5-8c3e-441b-868e-0a2e0ac0052c)
![Image](https://github.com/user-attachments/assets/c5002e08-80dd-4f75-beb9-86a65f494468)

deathbeam:
Are you using copilot.vim? And can you try logging out from it and then deleting json files in ~/.config/github-copilot?

ThomasCode92:
I use LazyVim with the [Copilot](https://www.lazyvim.org/extras/ai/copilot) and [Copilot Chat](https://www.lazyvim.org/extras/ai/copilot-chat) Extras,
I deleted the `~/.config/github-copilot` folder and logged back in, but got the same error.

ThomasCode92:
This is the output for `:CopilotChagModels` filtered by _github\_models_.

![Image](https://github.com/user-attachments/assets/59fd3860-0f38-4aa0-9703-8a4131c264e3)

deathbeam:
Alright so you have gpt-4o selected on github_models, disable the github_models provider or use model that is for regular copilot. Example on how to disable github models: https://github.com/deathbeam/dotfiles/blob/master/nvim/.config/nvim/lua/config/copilot.lua#L146

ThomasCode92:
@deathbeam Switching to a non _github\_models_ model (gpt-4o) works!
What is the difference between just _github_ anb _githug\_models_, if I may ask?

deathbeam:
> [@deathbeam](https://github.com/deathbeam) Switching to a non _github_models_ model (gpt-4o) works! What is the difference between just _github_ anb _github_models_, if I may ask?

[...]

PR #1061, opened 2025-03-27
fix(copilot)!: allow overriding headers, api_base

80avin:
We are using organizational version of Github copilot and I used to get unauthorized error, similar to https://github.com/CopilotC-Nvim/CopilotChat.nvim/issues/1052 ,

Digging deeper into the vscode's Copilot-chat extension, I found 2 differences which are required to fix this
1. API calls are made to `api.business.githubcopilot.com`
2. It doesn't work unless I set the headers as
   ```lua
   ["Editor-Version"] = "vscode/1.98.2",
   ["Editor-Plugin-Version"] = "copilot-chat/0.25.1",
   ```

As a solution, this PR implements
1. Use the base URL provided in auth response.
2. Allow overriding the headers.

Note: I haven't tested github_models because they weren't working under my credentials.

### DIFF
--- a/README.md
+++ b/README.md
@@ -417,10 +417,10 @@ Custom providers can implement these methods:
   embed?: string|function,
 
   -- Optional: Get extra request headers with optional expiration time
-  get_headers?(): table<string,string>, number?,
+  get_headers?(self: CopilotChat.Provider): table<string,string>, number?,
 
   -- Optional: Get API endpoint URL
-  get_url?(opts: CopilotChat.Provider.options): string,
+  get_url?(self: CopilotChat.Provider, opts: CopilotChat.Provider.options): string,
 
   -- Optional: Prepare request input
   prepare_input?(inputs: table<CopilotChat.Provider.input>, opts: CopilotChat.Provider.options): table,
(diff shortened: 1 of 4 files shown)


deathbeam:
But sending self might be fine but I dont think you need to add the changing api base anywhere else other than copilot provider, in embeddings simply use value from there when building the url and thats it and you dont need to do any changes to github_models. But def remove the default_headers, you are supposed to make your own provider when doing that (which is very simple you just reference the methods from your custom implementation and thats it right)

80avin:
@deathbeam, I agree with you.
I was also uncomfortable with the overriding of headers as it felt misplaced. The idea of custom provider didn't hit me.

Anyways, do you agree with keeping the api_base? A few points:
1. Github's API response itself tells us about using a different endpoint, and the official copilot-chat extension respects this. Hence, it sounds fair if we also do the same.
2. If api_base is a separate variable like current, I can essentially extend the copilot provider in my config to override `get_headers` and `api_base` instead of writing a new provider entirely.

deathbeam:
Well I dont want to keep api base as part of the default api, it should be copilot provider specific. you can always refer to it from other providers in the actual functions, and updating it from the inside of copilot provider. E.g its very specific to the copilot provider so it should just stay there imo.

80avin:
@deathbeam Can you review now ?

For others, the following lazy.nvim config shows how I'm able to override them.

```lua
{
    -- url = "https://github.com/80avin/CopilotChat.nvim",
    opts = function(opts)
      local providers = require("CopilotChat.config.providers")
      local orig_provider = providers.copilot
      ---@diagnostic disable-next-line: inject-field
      orig_provider._cache_get_header = orig_provider._cache_get_header or orig_provider.get_headers
      local orig_get_header = orig_provider._cache_get_header
      orig_provider.get_headers = function(self)
        ---@diagnostic disable-next-line: need-check-nil
        local headers, expires_at = orig_get_header(self)
        return vim.tbl_extend("force", headers, {
          ["Editor-Version"] = "vscode/1.98.2",
          ["Editor-Plugin-Version"] = "copilot-chat/0.25.1",
        }),
          expires_at
      end
      providers.github_models.disabled = true
      if true then
        return nil
      end
    end,
}
```

Issue #1071, opened 2025-03-28
scan_dir utility broken with ripgrep < 14


JPricey:
The `-d` short form for setting search max depth on ripgrep search was only added in version 14.0.  ([Changelog](https://github.com/BurntSushi/ripgrep/blob/master/CHANGELOG.md#1400-2023-11-26))
Making this plugin compatible for ripgrep < 14 is a simple matter of using the longform `--max-depth` instead.
I will have a PR up for this quick fix soon.

PR #1072, opened 2025-03-28
fix: use --max-depth arg name for ripgrep <14 compat

JPricey:
Closes https://github.com/CopilotC-Nvim/CopilotChat.nvim/issues/1071

I found this when trying to add a #files context, and I only have ripgrep 13.x installed on my system.


deathbeam:
thanks!

@all-contributors add @JPricey for code

PR #1087, opened 2025-04-01
feat: accept a function that returns the layout to use

ouuan:
This resolves #1080.

Example:

```lua
  window = {
    layout = function()
      if vim.o.columns > 2 * vim.o.lines then
        return 'vertical'
      else
        return 'horizontal'
      end
    end
  },
```

### DIFF
--- a/lua/CopilotChat/ui/chat.lua
+++ b/lua/CopilotChat/ui/chat.lua
@@ -33,8 +33,6 @@ local function match_header(header)
   end
 end
 
----@alias CopilotChat.ui.Chat.Layout 'vertical'|'horizontal'|'float'|'replace'
-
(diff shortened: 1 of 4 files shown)


lua/CopilotChat/ui/chat.lua
```diff
@@ -33,6 +33,8 @@ local function match_header(header)
   end
 end
 
+---@alias CopilotChat.ui.Chat.Layout 'vertical'|'horizontal'|'float'|'replace'
```
deathbeam, line 36: this should be in config

deathbeam (changes requested):

deathbeam:
Thanks!

@all-contributors add @ouuan for docs and code

PR #1029, opened 2025-03-17
feat(context)!: switch from contexts to function calling

deathbeam:
BEGIN_COMMIT_OVERRIDE
feat(context)!: switch from contexts to function calling

This change modernizes the CopilotChat architecture by implementing a tools-based 
approach using function calling. Each tool has a schema definition that enables 
structured parameter collection and validation. Resources are now handled more 
consistently, with a clearer distinction between content types.

https://platform.openai.com/docs/guides/function-calling?api-mode=responses
https://modelcontextprotocol.info/specification/2024-11-05/server/tools/

BREAKING CHANGE: Multiple breaking changes due to big refactor:
- The context API has changed from callback-based input handling to schema-based definitions.
- config.contexts renamed to config.tools
- config.context removed, use config.sticky
- diagnostics moved to separate tool call, selection and buffer calls no longer include them by default
- gi renamed to gc, now also includes selection
- filenames renamed to glob
- files removed (use glob together with tool calling instead, or buffers/quickfix)
- copilot extension agents removed, tools + mcp servers can replace this feature and maintaining them was pain, they can still be implemented via custom providers anyway
- actions and integrations action removed as they were deprecated for a while
- config.questionHeader, config.answerHeader moved to config.headers.user/config.headers.assistant

Closes #1045
Closes #1053 
Closes #1076 
Closes #1090
Closes #1096
Closes #526 

END_COMMIT_OVERRIDE

TODO:
- [x] Improve response format from contexts
- [x] Maybe rename contexts to tools?
- [x] Add names of tool to embedded context value
- [x] Inline contexts/tools - expand directly to user prompt
- [x] Split diagnostics to separate context
- [x] Mark built in tools as safe, mark everything else unsafe
- [x] How to resolve MCP resources? How to resolve built in resources? Currently hardcoded for file and url, not good
- [x] Rest of commands resources for reuse? git?
- [x] Properly store tool responses using "tool" role and also send tool_calls in history
  - When tool_call response is received, new tool_call prompt, user submitable, editable, sends only tools
  - On submit replace with tool_call responses, send proper "tool" role, wait response
- [ ] Parse resource template to schema for input, pass uri to input for resolution
- [x] Rename agent to group, allow sharing single tools as well 

https://github.com/ravitemer/mcphub.nvim integration:

```lua
local mcp = require('mcphub')
mcp.setup()
mcp.on({ 'servers_updated', 'tool_list_changed', 'resource_list_changed' }, function()
    local hub = mcp.get_hub_instance()
[...]

PR #1115, opened 2025-04-18
fix(validation): Ensure If the `terminal buffer` is excluded.

### LINKED ISSUE #1091: Exclude `Terminal buffer` from #buffers.
Either provide a way to exclude one or more files from `#buffers` or provide a way to exclude `terminal buffer` from the `#buffers`

Let's say I have 10 buffers open; one of them is `:term` buffer that is running my development server.

Now, when I use the `#buffers` command, it includes the terminal buffer, and this causes `token limit exceed`.
So, if there was a feature to exclude the terminal, then it would have been great. As I cannot close my dev server every time.

m4dd0c:
closes #1091 

### Problem
As I have explained, the current implementation of `utils.buf_valid` is not excluding `terminal buffer`, that is causing the token exceed problem.

### Solution

Added a condition to check if the buffer is `terminal_buffer` using `buftype`.
```lua
--- Check if a buffer is valid
--- Check if the buffer is not a terminal
---@param bufnr number? The buffer number
---@return boolean
function M.buf_valid(bufnr)
  return bufnr
      and vim.api.nvim_buf_is_valid(bufnr)
      and vim.api.nvim_buf_is_loaded(bufnr)
      and vim.bo[bufnr].buftype ~= 'terminal'
    or false
end
```

### DIFF
--- a/lua/CopilotChat/utils.lua
+++ b/lua/CopilotChat/utils.lua
@@ -210,10 +210,15 @@ function M.kv_list(tbl)
 end
 
 --- Check if a buffer is valid
+--- Check if the buffer is not a terminal
 ---@param bufnr number? The buffer number
 ---@return boolean
 function M.buf_valid(bufnr)
-  return bufnr and vim.api.nvim_buf_is_valid(bufnr) and vim.api.nvim_buf_is_loaded(bufnr) or false
+  return bufnr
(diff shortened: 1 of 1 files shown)


deathbeam:
Thanks. Dont think this is the best place for this check but it causes least conflicts with #1029 so for now its good enough.

@all-contributors add @m4dd0c for code

PR #1167, opened 2025-06-21
feat: introduce diagnostics context and prompt

gennaro-tedesco:
I often find myself wanting to resolve the code diagnostics and manually adding the corresponding error messages and/or buffers to the chat, together with a custom prompt. I thought I could propose to add diagnostics as default context and prompt that the user can select in order to streamline this process.

This PR introduces:

- context: workspace diagnostics for all open buffers
- prompts: instructions for the model to patch each single one independently


gennaro-tedesco:
Ah, very well, then we can close this PR I guess.

P. S. I don't see the `tools` branch (I wanted to install it) - is it the `canary` branch or which one?

PR #1170, opened 2025-06-22
fix: #1153 use filepath on accept

### LINKED ISSUE #1153: Accepting diffs with C-y strips away subdirectories
_Version:_

CopilotChat: Commit 16d897f
Neovim: 0.11.1

_Steps to reproduce:_

1. Create an empty directory
1. Create a `src` subdirectory within it
1. Create a `main.rs` inside the subdirectory created in step 2.
1. Add something to the file (I could not get CopilotChat to add anything to an empty file)
1. Open neovim in the empty directory from step 1. Tell CopilotChat (using sonnet 3.7 thinking in this case) to add something to `src/main.rs`
1. It should create a correct diff, something along the lines of [1] below
1. Accepting the diff with `C-y` inserts the diff into a new buffer named `main.rs`

_Expected behavior:_

I would expect the diff to be inserted into the `main.rs` file under `src/`

_Debugging:_

It seems like the changes added in commit 81754ea, addressing issue #1095, led to this behavior. Reverting the commit fixed the issue for me. Since I don't quite understand why normalization was initially added, I'm hesitant to create a pull request dealing with the issue.

[1] 
````
[file:src/main.rs](src/main.rs) line:1-1

```rust
 fn main() {
    println!("Hello, world! 🙂");
}
```
````

adborden:
Use filepath for the full path when accepting a change.

Fixes #1153

### DIFF
--- a/lua/CopilotChat/config/mappings.lua
+++ b/lua/CopilotChat/config/mappings.lua
@@ -64,7 +64,7 @@ local function get_diff(block)
     change = block.content,
     reference = reference or '',
     filetype = filetype or '',
-    filename = utils.filename(filename),
+    filename = utils.filepath(filename),
     start_line = start_line,
     end_line = end_line,
     bufnr = bufnr,


deathbeam:
@all-contributors add @adborden for code

PR #1174, opened 2025-06-23
feat: allow disabling context tip

SamSaffron:
In some cases context tip can lead llms to attempt tool usages by disabling
system prompt will no longer include context tips


deathbeam:
This is already a feature via:

chat.setup({
    include_contexts_in_prompt = false,
})

PR #1251, opened 2025-08-02
Slight fix for doc (CopilotChat.txt)

swnakamura:
Function name for git information has changed from `git:` to `gitdiff:`, but there's a line in documentation that uses the old function name. It is fixed.


deathbeam:
The .txt is generated from README so adjust it there if you can

deathbeam:
The readme and examples were reworked so this is no longer necessary, thanks anyway

PR #1263, opened 2025-08-03
fix: functions are missing from parse_schema's return

xavierchanth:
I'll use the `file` function's schema as an example of the problem:

```lua
    schema = {
      type = 'object',
      required = { 'path' },
      properties = {
        path = {
          type = 'string',
          description = 'Path to file to include in chat context.',
          enum = function(source)
            local glob = utils.glob(source.cwd(), {
              max_count = 0,
            })
            return glob
          end,
        },
      },
    }
```

At the end of `parse_schema`, the schema is filtered. After this gets called, only `type` and `description` are in the returned schema, but `enum` should be in there too.

```lua
  if schema then
    schema = filter_schema(schema, true)
  end

  return schema
```

Prior to this, completing `#file:` in the picker would give me an input field to type my file path:

image

This change restores the behavior visible in the demo in the README where that file can be picked using `vim.ui.select`:

image

### DIFF
--- a/lua/CopilotChat/functions.lua
+++ b/lua/CopilotChat/functions.lua
@@ -56,9 +56,7 @@ local function filter_schema(tbl, root)
   local result = {}
   for k, v in pairs(tbl) do
     if not utils.empty(v) then
-      if type(v) == 'function' then
-        result[k] = v
-      elseif k ~= 'examples' then
+      if k ~= 'examples' then
         result[k] = type(v) == 'table' and filter_schema(v) or v
       end
     end


deathbeam:
Good catch thanks. I fixed it myself because if I did this, the enum function would also be sent to the copilot API, and serializing functions as json is not pretty and it can break the API for some models, so the fix needed to be done only for when the input is being entered.

XavierChanth:
> Good catch thanks. I fixed it myself because if I did this, the enum function would also be sent to the copilot API, and serializing functions as json is not pretty and it can break the API for some models, so the fix needed to be done only for when the input is being entered.

Np, I had a feeling what I did wasn’t correct, but at least raising the PR identifies the issue. Thanks for the quick response!

PR #1318, opened 2025-08-12
Feat/dynamic system prompt

### LINKED ISSUE #1285: Use `.github/copilot-instructions.md` for system instructions.
Hi all,

The fact of having a specific system instruction for each workspace or project is something usefull, as **instructions change with the project**.

[Github has defined a location for system instructions](https://docs.github.com/en/copilot/how-tos/configure-custom-instructions/add-repository-instructions).

It would be a good idea to also use it in `CopilotChat.nvim`.

I submit a PR for that, feel free to discuss.

rakotomandimby:
Supersedes the previous PR that attemped to solve #1285 

With this modifications, if the "system instructions" are in the file `.github/instructions.md`, the user can configure `CopilotChat.nvim` this way:

```lua
    system_prompt = function()
      local file_path = vim.fn.getcwd() .. '/.github/instructions.md'
      if vim.fn.filereadable(file_path) == 0 then
        return "You are an AI programming assistant."
      end
      local file_content = vim.fn.readfile(file_path)
      return table.concat(file_content, '\n')
    end
```

Closes #1285

### DIFF
--- a/lua/CopilotChat/config.lua
+++ b/lua/CopilotChat/config.lua
@@ -125,4 +125,3 @@ return {
   -- default mappings
   mappings = require('CopilotChat.config.mappings'),
 }
-
--- a/lua/CopilotChat/init.lua
+++ b/lua/CopilotChat/init.lua
@@ -1148,4 +1148,3 @@ function M.setup(config)
 end
 
 return M
-


deathbeam:
Thanks!

@all-contributors add @rakotomandimby for code

PR #1332, opened 2025-08-15
handle * and ** more reliably

### LINKED ISSUE #1331: Make `glob_to_pattern` more reliable when it translates `*` and `**` wildcards
Hello,

The `utils.glob_to_pattern` function, which converts glob patterns to Lua regex for file searching, has a flaw in its implementation that leads to incorrect file matching.

`*` is translated incorrectly: The glob `*` is converted to the Lua pattern `.*`. This is incorrect because `*` should not match directory separators (`/`), but `.` does. This causes patterns like `*.lua` to incorrectly match files in subdirectories (e.g., `lua/utils.lua`).

`**` is not handled correctly: The glob `**` is not specially handled and is treated as two separate `*`s. This results in an incorrect pattern that does not properly match across multiple directory levels as intended.

I am working on a PR.

rakotomandimby:
Fix #1331


lua/CopilotChat/utils.lua
```diff
@@ -784,13 +784,19 @@ function M.split_lines(text)
 end
 
 --- Convert glob pattern to regex pattern
---- https://github.com/davidm/lua-glob-pattern/blob/master/lua/globtopattern.lua
 ---@param g string The glob pattern
 ---@return string
 function M.glob_to_pattern(g)
+  -- Handle ** patterns by preprocessing them
```
deathbeam, line 790: hmm maybe its good idea to add tests now for this function. we dont really have many tests but modifying this func is pretty scary, dont really want to break the pattern expansion

lua/CopilotChat/utils.lua
```diff
@@ -912,6 +918,13 @@ function M.glob_to_pattern(g)
       p = p .. escape(c)
     end
   end
+
+  -- Now handle the ** placeholder - replace with pattern that matches any path
+  -- including directory separators and zero directories
+  p = p:gsub(vim.pesc(DOUBLESTAR_PLACEHOLDER .. '/'), '.-/') -- **/something -> .-/something
+  p = p:gsub('/' .. vim.pesc(DOUBLESTAR_PLACEHOLDER), '/.-') -- something/** -> something/.-
```
deathbeam, line 925: hmm this doenst really make sense, if you gsub the doublestar_placeholder in the line above this, then this second gsub will never match because it was already replaced no?

deathbeam (changes requested):

rakotomandimby:
Ok @deathbeam , I will proceed to the required tests.
They will answer about the logic potential issues.

But do you agree on the relevance of the need of fixing  `glob_to_pattern` ?

rakotomandimby:
@deathbeam 

To install the test tooling, I had to install exactly `lua51`:

```
sudo pacman -S lua51
sudo luarocks --lua-version=5.1 install vusted
```

I use Arch BTW.
And it works

Tests passed:
```
20:19 $ make test
vusted test
ok 1 - CopilotChat plugin should be able to load
ok 2 - utils.glob_to_pattern should handle simple filenames
ok 3 - utils.glob_to_pattern should handle single asterisk for any characters except slash
ok 4 - utils.glob_to_pattern should handle question mark for a single character except slash
ok 5 - utils.glob_to_pattern should handle character sets
ok 6 - utils.glob_to_pattern should handle negated character sets
ok 7 - utils.glob_to_pattern should handle double asterisk for zero or more directories
1..7
```

deathbeam:
And extend the existing tests with yours pls

rakotomandimby:
I had to install `lpeg` in order to launch the tests (that is not a problem for me):
```
 sudo luarocks --lua-version=5.1 install lpeg
```

The last commit passes all tests:

```
nvim --headless --noplugin -u ./scripts/test.lua
Starting...Scheduling: tests/utils_spec.lua
Scheduling: tests/init_spec.lua
Scheduling: tests/functions_spec.lua
Error detected while processing /home/mihamina/.config/nvim/init.lua:
E5113: Error while calling lua chunk: /home/mihamina/.config/nvim/lua/options.lua:1: module 'nvchad.options' not found:
========================================	
Testing: 	/home/mihamina/Projects/RKTMB/rktmb-copilot-chat/tests/functions_spec.lua	
Success	||	CopilotChat.functions uri_to_url replaces parameters in uri template	
Success	||	CopilotChat.functions uri_to_url leaves missing params empty	
Success	||	CopilotChat.functions match_uri matches uri and extracts parameters	
Success	||	CopilotChat.functions match_uri returns nil for non-matching uri	
Success	||	CopilotChat.functions match_uri returns empty table for exact match with no params	
Success	||	CopilotChat.functions parse_schema returns schema if present	
Success	||	CopilotChat.functions parse_schema generates schema from uri if missing	
Success	||	CopilotChat.functions parse_input parses input string into table	
Success	||	CopilotChat.functions parse_input returns input if already table	
Success	||	CopilotChat.functions parse_input returns empty table if no schema	
	
Success: 	10	
Failed : 	0	
Errors : 	0	
========================================	
Testing: 	/home/mihamina/Projects/RKTMB/rktmb-copilot-chat/tests/utils_spec.lua	
Success	||	CopilotChat.utils glob_to_pattern: abc.*	
Success	||	CopilotChat.utils glob_to_pattern: file.lua	
Success	||	CopilotChat.utils glob_to_pattern: *.lua	
Success	||	CopilotChat.utils glob_to_pattern: src/*.lua	
Success	||	CopilotChat.utils glob_to_pattern: file?.lua	
Success	||	CopilotChat.utils glob_to_pattern: [abc].lua	
Success	||	CopilotChat.utils glob_to_pattern: [a-c].lua	
Success	||	CopilotChat.utils glob_to_pattern: [!abc].lua	
Success	||	CopilotChat.utils glob_to_pattern: [^abc].lua	
Success	||	CopilotChat.utils glob_to_pattern: **/*.lua	
Success	||	CopilotChat.utils glob_to_pattern: **/test/*.lua	
Success	||	CopilotChat.utils glob_to_pattern: src/**/test/*.lua	
Success	||	CopilotChat.utils glob_to_pattern: **	
Success	||	CopilotChat.utils empty	
Success	||	CopilotChat.utils split_lines	
Success	||	CopilotChat.utils make_string	
Success	||	CopilotChat.utils uuid	
Success	||	CopilotChat.utils to_table	
	
Success: 	18	
Failed : 	0	
Errors : 	0	
========================================	
Testing: 	/home/mihamina/Projects/RKTMB/rktmb-copilot-chat/tests/init_spec.lua	
Success	||	CopilotChat module should be able to load	
Success	||	CopilotChat module should be able to set up	
	
Success: 	2	
Failed : 	0	
Errors : 	0	
========================================	


```

deathbeam:
Hmm thats odd, i did not had to. but also i think i should've used --clean instead of --noplugin. Can you also revert the move of the function from utils, and also rebase or merge properly?

deathbeam:
> Ah, I pulled out the glob logic because I thought it would be massive over the time and the `utils.lua` approaches the thousand lines of code... so, I thought it is better to split.
> 
> Do you prefer a big `utils.lua` file?

well ideally dont do refactors while fixing bugs because that makes reviewing harder. also extracting the function to file thats just the function isnt the best refactor either. i want to split the utils at some point, but it would be to something like text/files/http etc

deathbeam:
> I am sorry, I don't understand what did you do and what do you want me to pull. Are there minor changes I have to make to my PR?

well check the changed files and the git diff, you did merge incorrectly i guess so now its all scuffed. i can try merging master again or just rebase

deathbeam:
ok merged the main here, now i can actually see the diff. but why did you changed all the test, just adding new test cases is enough

rakotomandimby:
@deathbeam ,

Regarding the test cases — if you don’t agree with my reasoning, I’m happy to revert, but I’d like to share my perspective first.

The original tests were focused on verifying the produced regex and the matching paths. That gave us a 3-step check: glob pattern → regex → paths. Technically, that’s valid, but the core goal of the feature is simply to match the paths.

My proposal is to simplify this to a 2-step check: glob pattern → paths. The generated regex would still be available for debugging if a test fails, but it wouldn’t be part of the core assertions.

If you’d prefer to keep the regex-based checks, I’ll gladly put them back — I just think this change makes the tests easier to read and maintain.

rakotomandimby:
Hello @deathbeam , 

I found small issues on the tests you want to be compliant with on  [test.lua#L29](https://github.com/davidm/lua-glob-pattern/blob/master/test.lua#L29):

```
-- basic * and ?
local p = globtopattern("abc.*")
assert(p == '^abc%..*$')
assert(("abc.txt"):match(p))
assert(("abc."):match(p))
assert(not("abc"):match(p))
```

This tests currently fails because the initial code expects `^abc%..*$` and the new code produces `^abc%.[^/]*$`
The `abc.*` glob should exclude slashes, though. but neither the produced regex nor the match cases tests dont consider slashes.

It would be a very big effort to really stick on the original specs.
Will you agree I slightly (no fundamental revamp) change initial test cases? Or do you prefer to stick with them?

PS: Note that on some cases, different variations of a regex can lead to the same result, so I am also asking for your validation to use variation if the new code produces a variation.

deathbeam:
> Hello @deathbeam ,
> 
> I found small issues on the tests you want to be compliant with on [test.lua#L29](https://github.com/davidm/lua-glob-pattern/blob/master/test.lua#L29):
> 
> ```
> -- basic * and ?
> local p = globtopattern("abc.*")
> assert(p == '^abc%..*$')
> assert(("abc.txt"):match(p))
> assert(("abc."):match(p))
> assert(not("abc"):match(p))
> ```
> 
> This tests currently fails because the initial code expects `^abc%..*$` and the new code produces `^abc%.[^/]*$` The `abc.*` glob should exclude slashes, though. But neither the produced regex nor the match cases tests dont consider slashes.
> 
> It would be a very big effort to really stick on the original specs. Will you agree I slightly (no fundamental revamp) change initial test cases? Or do you prefer to stick with them?
> 
> PS: Note that on some cases, different variations of a regex can lead to the same result, so I am also asking for your validation to use variation if the new code produces a variation.

For the tests that do not match, feel free to adjust them. The important part is that its clearly visible what changed, do not try to make the existing tests pass if they also contain the bug. The spec that vim.glob for example follows is this:

https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#pattern

but vim.glob is parsing the glob to lpeg instead of lua patterns (technically switching from plenary scan_dir_async to some simple alternative could also be an option i guess so we can use vim.glob instead)

deathbeam:
Actually after writing my last response I reaized it might just be easier to rewrite plenary scan_dir_async using lpeg for matching instead. So i did that here, then the glob parsing will also be covered by neovim tests:

https://github.com/CopilotC-Nvim/CopilotChat.nvim/pull/1339

So this should also hopefully fix #1331

PR #1340, opened 2025-08-17
feat(functions)!: remove selection and change it to resource function

deathbeam:
This change removes the automatic selection feature from CopilotChat, including all related code, configuration options, documentation, and highlighting. Selection management is now handled via resources and functions, simplifying the codebase and improving maintainability.


deathbeam:
Actually not sure if this is even good idea because selection is pretty useful for highlighting etc

PR #1347, opened 2025-08-21
fix(makefile): handle MSYS_NT as a valid Windows environment

### LINKED ISSUE #1345: Makefile errors with "Unsupported operating system" on Windows 11
I'm (unfortunately) having to use Windows 11 currently, and I'm getting an issue when installing CopilotChat.nvim via Lazy in Neovim.

The makefile says:

```
ifeq ($(UNAME), Linux)
    OS := linux
    EXT := so
else ifeq ($(UNAME), Darwin)
    OS := macOS
    EXT := dylib
else ifeq ($(UNAME), Windows_NT)
    OS := windows
    EXT := dll
else
    $(error Unsupported operating system: $(UNAME))
endif
```

But on my system, under Powershell, `uname` doesn't return `Windows_NT`, it returns `MSYS_NT-10.0-26100`

If I modify the OS check as follows, everything installs and works correctly:

```
ifeq ($(UNAME), Linux)
    OS := linux
    EXT := so
else ifeq ($(UNAME), Darwin)
    OS := macOS
    EXT := dylib
else ifeq ($(UNAME), Windows_NT)
    OS := windows
    EXT := dll
else ifneq ($(findstring MSYS_NT,$(UNAME)),)
    OS := windows
    EXT := dll
endif
```

Is this known about/documented anywhere?

AjmalShajahan:
Fixes #1345

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -10,6 +10,9 @@ else ifeq ($(UNAME), Darwin)
 else ifeq ($(UNAME), Windows_NT)
     OS := windows
     EXT := dll
+else ifneq ($(findstring MSYS_NT,$(UNAME)),)
+    OS := windows
+    EXT := dll
 else
     $(error Unsupported operating system: $(UNAME))
 endif


deathbeam:
@all-contributors add @AjmalShajahan for code

Issue #1350, opened 2025-08-23
Include diagnostics only from selection


ruicsh:
Hey. Could there be a new scope to the diagnostics function that only includes the ones found in the current selection? 

The use case would be: after selecting a function and prompting it to fix it, you want it to focus on the errors/warnings found on that function not on the whole file (`#diagnostics:current`). Something like `#diagnostics:selection` maybe?

deathbeam:
Fixed by #1351

PR #1351, opened 2025-08-23
feat(functions): add scope=selection to diagnostics

ruicsh:
Adds a new scope to the diagnostics function where the user specifies that only diagnostics found inside the visual selection are to be included in the prompt's context.

This implements #1350

### DIFF
--- a/lua/CopilotChat/config/functions.lua
+++ b/lua/CopilotChat/config/functions.lua
@@ -280,7 +280,7 @@ return {
         scope = {
           type = 'string',
           description = 'Scope of buffers to use for retrieving diagnostics.',
-          enum = { 'current', 'listed', 'visible' },
+          enum = { 'current', 'listed', 'visible', 'selection' },
           default = 'current',
         },
         severity = {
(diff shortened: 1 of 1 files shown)


deathbeam:
Ok #1340 was merged so now the selection is just select.get(source.bufnr)

ruicsh:
> Ok #1340 was merged so now the selection is just select.get(source.bufnr)

@deathbeam cheers, I updated the selection call

deathbeam (changes requested):

lua/CopilotChat/config/functions.lua
```diff
@@ -344,6 +344,19 @@ return {
...
       end
 
+      -- By default, collect from the whole buffer
+      local selection_start_line = 1
+      local selection_end_line = vim.api.nvim_buf_line_count(source.bufnr)
+      -- Determine selection range if scope is 'selection'
+      if scope == 'selection' then
+        local select = require('CopilotChat.select')
+        local selection = select.get(source)
+        if selection then
```
deathbeam, line 354: if selection is not found and its selected, this should return empty early

deathbeam:
awesome, thanks!

@all-contributors add @ruicsh for code

PR #1353, opened 2025-08-23
fix #1352: validate file readability and stops processing if failure

### LINKED ISSUE #1352: Stop prompt processing if referenced `#file` is not found (to save requests)
Currently, when a user provides a prompt along with one or more `#file` references, if any of the files are not found, the system still processes the prompt but ignores the missing file(s).

This forces the user to correct the file path and re-submit the prompt, consuming additional requests. Since requests are metered, this behavior wastes request quota.

A better approach would be to abort prompt processing when a referenced file is missing, so the user can immediately fix the path before retrying.

Example:
- User writes `./path/to/file.ext` instead of `path/to/file.ext` (the only difference being the leading `./`).
- Many other cases can also occur where a path is unintentionally wrong.

This change would help users avoid unnecessary request usage caused by missing file references.

rakotomandimby:
fix #1352: validate file readability and stops processing if failure

### DIFF
--- a/lua/CopilotChat/init.lua
+++ b/lua/CopilotChat/init.lua
@@ -374,6 +374,36 @@ function M.resolve_functions(prompt, config)
     })
   end
 
+  -- Validate file references before processing
+  local missing_files = {}
+  for _, pattern in ipairs(matches:keys()) do
+    if not utils.empty(pattern) then
+      local match = matches:get(pattern)
+      local name = match.word
+      local input = match.input
+
+      -- Check if this is a file function reference
+      local tool = M.config.functions[name]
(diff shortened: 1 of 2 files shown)


rakotomandimby:
@deathbeam , 
`fail_on_error` would be too much global, no?
Is it to be handle in this issue or will we create a new one for that extended scope?

deathbeam:
> @deathbeam , `fail_on_error` would be too much global, no? Is it to be handle in this issue or will we create a new one for that extended scope?

You are changing default behaviour, the config needs to be part of this PR to preserve it. It can just be named fail_on_tool_error or something. And by default it needs to be false

PR #1364, opened 2025-08-25
Fix #1352 : configuration parameter to stop on tool failure

### LINKED ISSUE #1352: Stop prompt processing if referenced `#file` is not found (to save requests)

rakotomandimby:
Fix #1352

### DIFF
--- a/lua/CopilotChat/config.lua
+++ b/lua/CopilotChat/config.lua
@@ -35,6 +35,7 @@
 ---@field auto_fold boolean?
 ---@field insert_at_end boolean?
 ---@field clear_chat_on_new_prompt boolean?
+---@field stop_on_tool_failure boolean?
 
 --- CopilotChat default configuration
 ---@class CopilotChat.config.Config : CopilotChat.config.Shared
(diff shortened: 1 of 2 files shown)


deathbeam:
Hmm actually I realized this should be stop_on_function_failure as it also bails on resource fails not only tool fails

deathbeam:
i will rename

rakotomandimby:
My bad. I will pay more attention on the future.

PR #1385, opened 2025-08-29
fix: Correct header highlighting for multi-byte characters

### LINKED ISSUE #1384: Header text partially highlighted
When using custom headers with emojis (e.g., `"👤 User "`), parts of the header text are incorrectly highlighted with different colors instead of using the consistent
  `CopilotChatHeader` highlight group.

  ## Steps to Reproduce
  1. Configure custom headers with emojis:
     ```lua
     require("CopilotChat").setup({
       headers = {
         user = "👤 User ",
         assistant = "🤖 Assistant ",
         tool = "🔧 Tool ",
       },
     })
  2. Open CopilotChat and start a conversation
  3. Observe the header text

  Expected Behavior

  The entire header text should use the CopilotChatHeader highlight group consistently.

  Actual Behavior

  Parts of the header text (e.g., "Us" in "👤 User") are highlighted with different colors (often orange/Function color), while other parts use the correct header color or remain
  uncolored.

<img width="848" height="68" alt="Image" src="https://github.com/user-attachments/assets/48e8e77e-1986-4be0-9d8c-7d6ae4701303" />

ctchen222:
The previous implementation used `vim.fn.strwidth()` to calculate the end column for header highlighting. This function returns the display width, which can differ from the byte length when multi-byte characters like emojis are present.

Since the `end_col` for extmarks expects a byte-based index, this caused the highlighting to be applied incorrectly.
 
This patch corrects the issue by using the byte length (`#header_value`) for the `end_col` of the highlight extmark, while still using `vim.fn.strwidth()` for `virt_text_win_col` to ensure correct visual alignment of UI elements.

Fixes #1384

### DIFF
--- a/lua/CopilotChat/ui/chat.lua
+++ b/lua/CopilotChat/ui/chat.lua
@@ -620,19 +620,19 @@ function Chat:render()
       if id then
         -- Draw the separator as virtual text over the header line, hiding the id and anything after the header
         if self.config.highlight_headers then
-          local sep_col = vim.fn.strwidth(header_value)
-          vim.api.nvim_buf_set_extmark(self.bufnr, highlight_ns, l - 1, sep_col, {
+          local header_width = vim.fn.strwidth(header_value)
+          vim.api.nvim_buf_set_extmark(self.bufnr, highlight_ns, l - 1, 0, {
+            end_col = #header_value,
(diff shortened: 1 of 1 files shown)


ctchen222:
Fix #1384

PR #1472, opened 2025-10-23
fix(chat): reinsert stickies before prompt processing

### LINKED ISSUE #1471: Sticky prompts are broken
My sticky `> #buffer` seems to have stopped working after 1ff0bb3b28d92e916d7c63a9189415ea2aaa24de. It still works if I provide a non-sticky (`#buffer`) which seems to have the buffer number appended automatically.

deathbeam:
This change ensures that sticky messages are properly reinserted before processing the prompt, preserving their context in the chat flow. The previous logic for removing sticky prefixes was removed, and stickies are now stored correctly after resolution.

Closes #1471


tghosgor:
Yeah it works fine after the PR. Thanks.

PR #1518, opened 2026-01-31
Support "Auto" model mode (Smart Model Selection) similar to VS Code 

### LINKED ISSUE #1517: Support "Auto" model mode (Smart Model Selection) similar to VS Code
The latest [Copilot Chat in VS Code](https://docs.github.com/en/copilot/concepts/auto-model-selection) extension has an update for the auto model selection feature. Its behavior can be simply summarized as:
- First, the user selects the "Auto" model from the model list
- The Copilot extension sends a POST request to the `/models/session` endpoint, which returns fields such as `available_models`, `selected_model`, etc. Among these, the `selected_model` field is an output based on intelligent model selection using current discount information, traffic, etc. (As of [Commit 7c4e32b](https://github.com/microsoft/vscode-copilot-chat/commit/7c4e32b310678c83524e994295ef04ac97fae893), the architecture based on context and routing APIs is still experimental and subject to change.)
- Next, the extension constructs the `completion` request based on the output of the intelligent model selection to complete the upload of the chat history.

You can refer to the following image for understanding (note: the screenshot was taken using Visual Studio Code Insider edition along with the Copilot extension built from the latest version, which may differ from the interface of the stable release)
Image

To update this plugin by adding auto model selection functionality, I envision that we could achieve this by:
- Reusing the `$` prefix and adding `$auto` to the model selection (it can also be set in the config via `model = "auto"`)
- When selecting the `auto` model, we construct a simulated request to mimic the above request and obtain the routed model's response.

pxwg:
Closes #1517.

### DIFF
--- a/lua/CopilotChat/client.lua
+++ b/lua/CopilotChat/client.lua
@@ -63,7 +63,6 @@ local class = require('CopilotChat.utils.class')
 local files = require('CopilotChat.utils.files')
 local orderedmap = require('CopilotChat.utils.orderedmap')
 local stringbuffer = require('CopilotChat.utils.stringbuffer')
-local model = require('fidget.notification.model')
 
 --- Constants
 local RESOURCE_SHORT_FORMAT = '# %s\n```%s start_line=% end_line=%s\n%s\n```'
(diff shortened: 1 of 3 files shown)


pxwg (commented):
I think this PR has served its purpose.

PR #1556, opened 2026-04-16
feat(trusted-tools): auto-execute trusted tool calls

deathbeam:
Add `trusted_tools` config option to allow automatic execution of
trusted tool calls without manual approval. By default, all tool calls
require approval, but users can now trust specific tools, groups, or all
tools. Update README and type annotations to document the new behavior.
Trusted tools are determined by function definition, group, or name.

Closes #1534


josealvaradoo:
thank you so much @deathbeam

PR #1568, opened 2026-06-01
fix: use `gpt-5-mini` as default model

RoseSecurity:
## what

Updated CopilotChat’s default model from `gpt-4.1` to `gpt-5-mini`.

Also updated the README and generated vimdoc examples so documented defaults and model examples match the new config value.

## why

GitHub Copilot lists GPT-4.1 as closing down on June 1, 2026. `gpt-5-mini` is the closest replacement because GitHub describes it as a general-purpose model suitable for fast, accurate coding and explanation tasks, matching the previous GPT-4.1 default use case.

## references

- GitHub Copilot supported models: https://docs.github.com/en/copilot/reference/ai-models/supported-models
- GitHub Copilot model comparison: https://docs.github.com/en/copilot/reference/ai-models/model-comparison

### DIFF
--- a/README.md
+++ b/README.md
@@ -240,7 +240,7 @@ Most users only need to configure a few options:
 
 ```lua
 {
-  model = 'gpt-4.1',           -- AI model to use
+  model = 'gpt-5-mini',        -- AI model to use
   temperature = 0.1,           -- Lower = focused, higher = creative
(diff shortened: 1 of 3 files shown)


RoseSecurity:
@deathbeam The default option is not available as of today

deathbeam:
Thanks. @all-contributors add @RoseSecurity for docs and code

PR #1570, opened 2026-06-02
fix(chat): render on `InsertEnter` when auto_insert_mode is enabled

### LINKED ISSUE #1564: Bug: Chat doesn't render when auto_insert_mode is set true until mode is switched to normal
There seems to be a bit of rendering issue when the chat buffer is opened in insert mode using the `auto_insert_mode` option.

From what I can tell, currently rendering is triggered on the events `TextChanged` and `InsertLeave` only, and so the render doesn't happen until leaving insert mode.

I may be missing something or may have misconfigured something, can anyone else reproduce this behavior?

Adding `InsertEnter` to the autocmd when `auto_insert_mode` is enabled seems to fix the issue locally, and I'd be happy to open a PR if this approach is acceptable.

abhr-0:
Closes: #1564

### DIFF
--- a/lua/CopilotChat/ui/chat.lua
+++ b/lua/CopilotChat/ui/chat.lua
@@ -610,7 +610,7 @@ function Chat:create()
     end
   end)
 
-  vim.api.nvim_create_autocmd({ 'TextChanged', 'InsertLeave' }, {
+  vim.api.nvim_create_autocmd({ 'TextChanged', 'InsertEnter', 'InsertLeave' }, {
     buffer = bufnr,
     callback = function()
       utils.debounce('chat-parse-' .. bufnr, function()


lua/CopilotChat/ui/chat.lua
```diff
@@ -610,7 +610,12 @@ function Chat:create()
     end
   end)
 
-  vim.api.nvim_create_autocmd({ 'TextChanged', 'InsertLeave' }, {
+  local render_events = { 'TextChanged', 'InsertLeave' }
+  if self.config.auto_insert_mode then
```
deathbeam: just always register InsertEnter too. auto_insert_mode can change during runtime, doing this conditionally is bad

deathbeam (changes requested):

lua/CopilotChat/ui/chat.lua
abhr-0: Makes sense, thanks. I've updated the PR accordingly.

deathbeam (approved):

deathbeam:
@all-contributors add @abhr-0 for code

